PR #512, opened 2015-01-18
Fixes #479

### LINKED ISSUE #479: Modify the [affiliate_referral_url] shortcode to show other URL variations
The [affiliate_referral_url] currently only outputs the referral URL in the following format:

`site.com/page/?ref=123`

We should modify the shortcode so it can show the other URL variations such as:

`site.com/page/ref/username`

We could implement a default referral URL setting in the admin and then the shortcode could output based on what is selected. Here's where the setting was discussed: https://github.com/affiliatewp/AffiliateWP/issues/421#issuecomment-59323983

sumobi:
@pippinsplugins I made 2 extra parameters, `pretty` and `format`. Format is used to either select `id` or `username`. When `pretty` is set to `yes` it shows a pretty URL. Between those 2 parameters we can have al the different variations of the referral URL.

### DIFF
--- a/includes/admin/settings/class-settings.php
+++ b/includes/admin/settings/class-settings.php
@@ -166,6 +166,10 @@ class Affiliate_WP_Settings {
 	*/
 	function get_registered_settings() {
 
+		// get currently logged in username
+		$user_info = get_userdata( get_current_user_id() );
+		$username  = esc_html( $user_info->user_login );
+
 		$settings = array(
 			/** General Settings */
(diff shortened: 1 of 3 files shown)


sumobi:
@pippinsplugins Going to also re-do the `dashboard-tab-urls.php` file in this PR to take into account the affiliate referral URL format etc.

sumobi:
@pippinsplugins Going to resubmit this as a branch.

Issue #506, opened 2015-01-15
Filter needed for updating affiliate


sumobi:
I need a filter when an affiliate is updated, right before the `$args` are sent to the `affiliate_wp()->affiliates->update()` function. 

Here is preferable:
https://github.com/affiliatewp/AffiliateWP/blob/master/includes/affiliate-functions.php#L725

    $args = apply_filters( 'affwp_update_affiliate_args', $args, $data );    

Any objections @pippinsplugins?

pippinsplugins:
We do need a filter but not there. We should just add a filter to the
`update()` method in the main DB class. That way we "magically" have an
update filter for all DB classes.

sumobi:
Yeah that's a good idea, let me have a look for a better position

sumobi:
I'm going to need the `$data` array passed into the update() method for this to work. 

Because I've added new fields to the affiliate screen I can only access them in `$data`, which is passed into `affwp_update_affiliate( $data = array )`.

See: https://github.com/affiliatewp/AffiliateWP/blob/master/includes/affiliate-functions.php#L709

So I'm guessing we pass `$data` into https://github.com/affiliatewp/AffiliateWP/blob/master/includes/affiliate-functions.php#L726?

sumobi:
It doesn't feel right passing $data into the update() method

sumobi:
Heh, I just realized that because I'm not storing my data in the custom WP table anymore (as per https://github.com/affiliatewp/AffiliateWP/issues/507) I won't be needing this filter.

@pippinsplugins I guess all I need is a `do_action()` hook. How about a do_action in the affwp_update_affiliate() so I can grab the data and chuck it into the usermeta table?

sumobi:
@pippinsplugins Thoughts on a hook similar to the `affwp_post_insert_$type` one in the insert() function?

    do_action( 'affwp_post_update', $data );

This would go right here:

https://github.com/affiliatewp/AffiliateWP/blob/master/includes/class-db.php#L122

Do we want to add the `$type` variable like https://github.com/affiliatewp/AffiliateWP/blob/master/includes/class-db.php#L87?

pippinsplugins:
Seems like a good idea to me!

sumobi:
do we need the $type variable? i'll leave off for now unless you can see a good reason for it

pippinsplugins:
Yep, we need `$type` so that we know what kind of data is being updated. How about we pass it as a fourth parameter.

PR #515, opened 2015-01-20
Add affwp_post_update() hook

sumobi:
Fixes: https://github.com/affiliatewp/AffiliateWP/issues/506


sumobi:
@pippinsplugins if we can slip this in 1.5.2 that would be awesome

pippinsplugins:
Yep, lets do it. Hoping to review / finish 1.5.2 tonight or tomorrow.

pippinsplugins:
Since this will fire on all update calls, did you have a method in mind to distinguish the data type being updated?

sumobi:
@pippinsplugins I don't yet, no. I was thinking we'd do the same as the add hook, but we'd need to pass in a type variable.

Issue #387, opened 2014-09-17
Add filter to enable affiliates to receive commissions on their own purchases


sumobi:
We'll add a filter to each integration's conditional so an affiliate can receive commission on their own referral:

    if( $this->get_affiliate_email() == $customer_email ) {
        return; // Customers cannot refer themselves
    }

pippinsplugins:
With #519 done, we can now introduce a simpler filter in the `is_affiliate_email()`

PR #516, opened 2015-01-20
First pass at #387 for EDD

sumobi:
#387

@pippinsplugins Look alright to you?


sumobi:
@pippinsplugins See if that's more like it

pippinsplugins:
We should make this a lot more flexible and make sure the product ID, the order ID, the affiliate ID, etc, are all available in the filter.

pippinsplugins:
Going at this a different way.

Issue #537, opened 2015-02-10
Add affiliate meta table to store custom affiliate data


ryanduff:
Should work similar to user meta, but allow direct lookup via affiliate ID instead of having to get user ID and get user meta.

ryanduff:
Sidebar: I see there are unit tests, are they currently working? If so I'll write up a few to test the meta functions.

pippinsplugins:
Woot!

Yes, unit tests are working and are currently run through Travis.ci, we just haven't written many yet. I'd _love_ to  have tests with this.

pippinsplugins:
Let's get this baby into 1.6.

pippinsplugins:
Proposal in #538 is fantastic. Love it.

ryanduff:
Thoughts on also adding wrapper functions like `affwp_get_affiliate_meta()` that basically trigger `affiliate_wp()->affiliate_meta->get_meta()` ?

Personally on the fence. But they do make it a bit simpler to remember like `get_post_meta()` and `get_user_meta()`

pippinsplugins:
:+1: to adding wrapper gunctions.

pippinsplugins:
Am I right that this would also support `get_metadata( 'affiliate', $object_id, $meta_key, $single )`?

ryanduff:
Yes. The functions in the meta db class just wrap that. 

The key to pulling it off is registering the table with `$wpdb` to the table check passes and it doesn't short out. 

https://github.com/ryanduff/AffiliateWP/blob/d6b74cc72c0076259835ffa956dcdd9b398c8aac/includes/class-affiliate-meta-db.php#L33-36

pippinsplugins:
Super cool.

pippinsplugins:
Now in `release/1.6`

PR #538, opened 2015-02-10
Affiliate meta table

ryanduff:
First pass at affiliate meta table for #537


pippinsplugins:
I've gone ahead and merged this manually into `issue/537`

PR #547, opened 2015-02-26
Fix missing closing bracket in integrations/class-woocommerce.php

ryanduff:
In e8c76ade0913ed8fa80775ac3c5fba546d468b95 a closing bracket was lost on the new span. This fixes the currently broken markup.


pippinsplugins:
Thanks!

Issue #392, opened 2014-09-29
Introduce emails class and template files


pippinsplugins:
Just as we recently did for EDD, we should introduce an `AffWP_Emails` class that

ghost1227:
I've got the main class pretty much built, can you tell me what initial template tags you want to support by core?

pippinsplugins:
Don't worry about template tags yet. We will have to decide how best to
handle those since they are not the same as the Easy Digitial Downloads
template tags. We will have several types of tags here.

pippinsplugins:
@sumobi This is ready for testing.

sumobi:
:+1: aye aye captain

PR #558, opened 2015-03-05
Add emails class

ghost1227:
Issue #392

This is NOT ready to be merged. This is solely being posted so I can get feedback from @pippinsplugins on a few things. IE: I'm currently instantiating a second email object (`$email` vs `$emails`) to maintain backwards compatibility. Is that ok in your book? Also, all previous classes are created as `Affiliate_WP_x`, but the issue indicated the new class be `AffWP_Emails`... was that your intention or do you prefer to stick with the original scheme? If we stick with the original scheme, we're conflicting with the existing emails class. Thoughts?


pippinsplugins:
It should use `Affiliate_WP_Emails` as the class name.

pippinsplugins:
Your new class needs to be merged with the existing emails class. They are to be the same class.

That also means there is no need for a second instance.

pippinsplugins:
Note: this also needs to add the template files and necessary settings.

ghost1227:
Already got much of the settings/template written. Just wanted clarification on the conflicts with the old class.

ghost1227:
@pippinsplugins Is this looking better so far?

pippinsplugins:
Looking good.

Would you be interesting in combining email tags and emails into a single class instead of two separate ones? It was done that way in EDD because email tags came first. Since we're doing them at the same time, we can combine them.

ghost1227:
Personally, I think it makes sense to keep them separate. Emails handles the actual construction and sending of the email, tags is about parsing email tags specifically. That said... it's your show.

pippinsplugins:
Andrew and I just chatted on it: one class please.

ghost1227:
@pippinsplugins how's that?

pippinsplugins:
Initially looks good. I'll have to play with it to say for sure.

One thing I know we'll need is support for affiliate ID and referral IDs in the email tag callbacks, but it wouldn't be for all callbacks, just some.

ghost1227:
I might need a bit of help. I've migrated the actions to the new structure, but I'm starting to get confused looking at the notification function itself. Not like I don't understand what it's doing... just having trouble focusing.

pippinsplugins:
Walk me through what part doesn't make sense and I"ll be happy to help.

ghost1227:
All we have left to convert is merging the old notification function into the new class, and finishing up templates and default tags.

pippinsplugins:
The only thing you need to do with the existing notifications is make it so
that they use the new email templates.

ghost1227:
Are you referring to the notification function? https://github.com/affiliatewp/AffiliateWP/blob/master/includes/class-emails.php#L52

pippinsplugins:
That should be able to remain unchanged. Just updated the send() function:
https://github.com/affiliatewp/AffiliateWP/blob/master/includes/class-emails.php#L120

ghost1227:
@pippinsplugins too late... I merged it into the new class with a few small tweaks. Hopefully you don't object to the tweaks (mostly pulling data from settings so things aren't hardcoded).

pippinsplugins:
No worries, we can always adjust if needed. So with that, what questions or
issues do you have?

ghost1227:
@pippinsplugins Two questions at the moment... first, why aren't those tags getting added/displayed correctly? I'm missing something, but I can't figure out what. Second, not all tags are relevant to all emails. Notably, the amount tag is only valid on new referral emails. How do you want to handle displaying said tags in the settings page?

pippinsplugins:
I really don't have an answer for how to handle the different types of tags (affiliates vs referrals). I'll think more on that.

Are none of the tags working? What happens with them? Just blank?

Looks like the actual template files have not been committed: https://github.com/affiliatewp/AffiliateWP/pull/558/files - Can you make sure you've added them?

ghost1227:
Haven't tested tag functionality, but display is returning nothing. It seems that the variable isn't being set somewhere, but I can't track down the bug.

screen shot 2015-03-09 at 8 11 14 pm

pippinsplugins:
have you stepped through line by line and var_dump()'ed it?

Also, see my note about the template files?

ghost1227:
I saw your note, I'm still trying to figure out how to handle template files. In EDD, we only have one template file (for receipts). In AffWP, we have three potential emails we need to work with. Do we have preview/test actions for all three? Allow selection of template for all three? Should any be text-only like the EDD admin notices are now? Do we do logos per-email, or just assume all use the same logo? There's a LOT more complexity in supporting three templates than one.

pippinsplugins:
It's exactly the same. There are three files: header, body, footer.

pippinsplugins:
Sorry one difference: no need for template select in settings

ghost1227:
So we're only supporting one template? Not allowing HTML vs text? What about logos?

pippinsplugins:
Correct: one HTML template only. Yes to logso.

ghost1227:
Ok... In EDD, header/body/footer are all in one file (templates.php). You said three files. Do you want them split out in AffWP?

pippinsplugins:
Three files:
https://github.com/easydigitaldownloads/Easy-Digital-Downloads/tree/master/templates/emails
(look at -default)

sumobi:
@ghost1227 just testing now. So far:

1. Typo

    PHP Fatal error:  Call to undefined function apply_filter() in /wp-content/plugins/affiliatewp-master/includes/emails/class-affwp-emails.php on line 165

2. This would have come over from EDD, seen this one before.

PHP Notice:  wp_editor() was called with an argument that is <strong>deprecated</strong> since version 3.9! TinyMCE editor IDs cannot have brackets.

sumobi:
Typo: "Contratulations"

ghost1227:
@sumobi All those fixed...

ghost1227:
The amount tag, being the only tag that applies JUST to the affiliate notification email, requires a parameter that no other tag does. I'm not sure how to pass it. This also screws up the html list output on the settings page.. amount lists on all emails, though it only applies to one. Not sure the best way to handle it.

pippinsplugins:
Resubmit to `issue/392` please.

Issue #562, opened 2015-03-06
Filter the creatives arguments


seb86:
When adding a new creative or editing a creative, there is no filter applied to the arguments.

pippinsplugins:
There are several actions / filters you can user actually.

Could you tell me what you're trying to do? I can point you in the right direction (assuming this is a filter).

seb86:
What I would like to do is create a custom field for my client that allows him to place a custom embed code rather than a text link or image. I put these filters in place so that the custom field data is passed through. Was I wrong to do this?

pippinsplugins:
No, not wrong, it's just unnecessary.

You can use the `affwp_insert_creative` hook to add custom data to a creative.

pippinsplugins:
Closing as wontfix since the affwp_insert_creative hook can be used.

PR #563, opened 2015-03-06
Filtered creative database add and update

seb86:
https://github.com/affiliatewp/AffiliateWP/issues/562


pippinsplugins:
See comment on #562

PR #576, opened 2015-03-18
Integration hooks

dancameron:
actions aren't technically necessary but when I worked on my integration I was looking for an easy way to load up the Affiliate_WP_Base subclass at the ideal time, which I could not find. This is different than filtering the integrations via affwp_enabled_integrations becuase there's no way to have the loop load up the subclass since the file path is a constant (and defined).


pippinsplugins:
Nice, looks good.

We'll merge once 1.6 gets closer.

Issue #498, opened 2015-01-08
Paid Memberships Pro - Per-membership commission amounts


sumobi:
Like per-product commission amounts, but for membership levels

codelion7:
I'm about to implement this myself, but I'm wondering if it's best to add the settings for each membership level within PMP's "Edit Membership Level" screen, or to simply add the settings to the "Integrations" tab  of AffiliateWP. What do you guys think is best?

pippinsplugins:
If vote for within the membership level edit screen so it is consistent
with our other integrations.

sumobi:
:+1:

codelion7:
Alright, will do.

pippinsplugins:
#670

pippinsplugins:
@codelion7 do you think you'll have the PR for this updated in the next day or two?

codelion7:
@pippinsplugins, I honestly don't think that I'll get to finishing this until the weekend. I'm going to be travelling tomorrow, and then, I'll be finishing MLM version 1.0.5 for the next few days.

pippinsplugins:
Ok I'll try and extract them from the previous PR.

pippinsplugins:
Moving to 1.8

michaelbeil:
Do we want to keep this in 1.8?

pippinsplugins:
Let's try

michaelbeil:
Ok. It looks like we could parse through some of the previously referenced PRs and add this option to the PMP membership level screen fairly easily.

DrewAPicture:
Added a PR: #1226

pippinsplugins:
@DrewAPicture Looking good overall. What's the purpose of the `product_type` parameter? Does not look like that should be required since we can just consider PMP levels to be products (similar to how they are treated in MemberPress and Restrict Content Pro integrations).

DrewAPicture:
@pippinsplugins The problem is lack of context. Injecting a new referral rate in between the global rate and the affiliate rate is a little tricky without completely overriding `calculate_referral_amount()` (which I looked at, but it was a huge rabbit hole, and raises the future maintenance burden).

I found that overriding `get_product_rate()` was the best option for injecting the new referral rate, mostly because a membership is a product, but in this case we don't pass a product ID and it doesn't use posts or post meta.

The 'product_type' parameter was added to allow selectively triggering `get_product_rate()` (which can then be overridden to work around the lack of posts and post meta).

[...]

Issue #556, opened 2015-03-05
Add products column to referrals table


pippinsplugins:
We need to add a `products` column to the referrals DB to hold the details of all products purchased and their respective referral rates.

I propose we store product details as a serialized array with the following keys:
- name
- price
- referral_amount
- id

This will allow us to track and look up how much of the referral total each product in a purchase is responsible for. With this information, store owners will be able to refund one product in a purchase and then properly adjust the commission total to reflect the refund. Right now it's all or nothing.

This change is required for us to fix #555

codelion7:
This is an excellent idea, and I think it will be perfect for my next project. I'm going to be expanding the PMP integration to allow user's to set Per Membership Level, Per Product Commission Rates.So for example, site admins will be able to set a different commission rate for the sale of Level 1 Memberships, based on the Membership Level of the referring affiliate.

pippinsplugins:
:+1:

ryanduff:
Going to try and get this one wrapped up and submit a PR submitted in the next few days. May also kill the need for #531 by having a dedicated products column.

pippinsplugins:
Awesome, thanks @ryanduff !

pippinsplugins:
@ryanduff I'm working on this now. Sorry if you'd already done some work on it.

pippinsplugins:
For 1.6, I'm adding this to all integrations that support per-product rates, which means WooCommerce, EDD, and Exchange.

pippinsplugins:
The products of a referral can be set like this:
``php
$referral_id = affiliate_wp()->referrals->add( 
	array(
		'amount'       => $referral_total,
		'reference'    => $payment_id,
		'description'  => $this->get_referral_description( $payment_id ),
		'affiliate_id' => $this->affiliate_id,
		'context'      => $this->context,
		'products'     => $this->get_products( $payment_id )
	)
);
```

Retrieving the products is as simple as accessing the `products` property of the referral object returned by `affwp_get_referral()`.

PR #586, opened 2015-03-22
Add Per-Membership Rates to Release/1.6

codelion7:
Resolves: #498


codelion7:
After I added Per-Membership Rates, I realized that the integrations for WooCommerce and Easy Digital Downloads did not allow site admins to set a Rate Type for a specific Product or Download. 

Since I have already done this for the PMP integration, I decided to go ahead and add this feature to these other 2 integrations as well. This way, site admins can set a flat commission rate for one product/download, and a percentage commission rate for another product/download.

includes/referral-functions.php
```diff
@@ -159,32 +159,39 @@ function affwp_delete_referral( $referral ) {
 	return false;
 }
 
-function affwp_calc_referral_amount( $amount = '', $affiliate_id = 0, $reference = 0, $rate = '', $product_id = 0 ) {
-
+function affwp_calc_referral_amount( $amount = '', $affiliate_id = 0, $reference = 0, $rate = '', $type = '', $product_id = 0 ) {
```
pippinsplugins: `$type` will need to be added as the last parameter since we don't know how other plugins may be using this function. Changing the order of the parameters will break calcualtions.

codelion7:
I did check through the repo for any reference to affwp_calc_referral_amount() just to make sure that it wouldn't affect anything else, but I do agree with you. Better safe than sorry.

pippinsplugins:
We can always change it in core but we don't know how other extensions are using the functions, so we have to be safe and not change it.

pippinsplugins:
I'll review this further and get it merged in for 1.6.

Thanks for all of your work on it!

includes/integrations/class-base.php
```diff
@@ -205,23 +206,37 @@ public function calculate_referral_amount( $base_amount = '', $reference = '', $
...
 	*/
 	public function get_product_rate( $product_id = 0, $args = array() ) {
-
-		$rate = get_post_meta( $product_id, '_affwp_' . $this->context . '_product_rate', true );
+		
+		// Get product-specific rate
+		$product_rate = get_post_meta( $product_id, '_affwp_' . $this->context . '_product_rate', true );
+		
+		// Get membership-specific rate
+		$membership_rate = get_option( 'affwp_' . $this->context . '_membership_rate_' . $product_id );
```
pippinsplugins: The membership rate should be the same thing as the product rate. For membership plugins, the `product` is the same as the `membership level`. No need to have two different meta keys here.

includes/integrations/class-base.php
```diff
@@ -231,5 +246,40 @@ public function get_product_rate( $product_id = 0, $args = array() ) {
...
+	 * @since   1.6
+	 * @return  string
+	*/
+	public function get_product_rate_type( $product_id = 0, $args = array() ) {
+		
+		// Get product-specific rate type
+		$product_rate_type = get_post_meta( $product_id, '_affwp_' . $this->context . '_product_rate_type', true );
+		
+		// Get membership-specific rate type
+		$membership_rate_type = get_option( 'affwp_' . $this->context . '_membership_rate_type_' . $product_id );
```
pippinsplugins: Same thing here. There's no need to have separate keys for membership and products; they're the same thing.

codelion7: I do agree with you on this, but when storing the product rate for a particular membership, I realized that in the case of PMP, each membership is not a post type and there was no way for me to store the data as post meta. The 2nd meta key is really just an alternate way of retrieving the product rate.

I'm going to revise the code so that the 2nd meta key (and the variable it's stored in) adheres as closely as possible to the original meta key format. I will still need to append the product id on the end for differentiation though.

pippinsplugins:
@codelion7 it's fine to not store it in meta but I think the key should be the same. Since this method could carry over to eCommerce plugins as well, it makes sense to keep the same naming scheme, whether in post meta or options.

For PMP, just overwrite get_product_rate/_type() in the PMP class instead of using the base class version.

Does that make sense?

codelion7:
Yes, that makes sense. I'll make the change.Also, what's the status of the Products column for the referrals table? Should I incorporate it into my per-product rates implementation?

pippinsplugins:
The `products` column is finished https://github.com/affiliatewp/AffiliateWP/issues/556 :)

codelion7:
Very nice. 

And just to clarify, you think it's best to create a PMP-specific version of get_product_rate/_type() in the PMP class, but should I remove the get_option() call from the base class as well?

There may not be a need for this alternate method of acquiring the product rate in other integrations if there's no way to specify which product the rate is set for.

pippinsplugins:
Yes, create methods in the class.

The base class should not use get_option().

pippinsplugins:
@codelion7 Looks like we'll need to punt this to 1.6.1/2/3 since we're just now finalizing the release for tomorrow.

Once those tweaks are done, could you resubmit the PR? You can submit to Master this time since it'll go out right away once completed.

codelion7:
Ok, understood. I'll work on it locally for now, and resubmit the PR once the Master branch is updated to 1.6.

pippinsplugins:
Great, thanks!

On Wed, Apr 8, 2015 at 11:26 AM, Christian Freeman <notifications@github.com

pippinsplugins:
@codelion7 Could you resubmit this soon?

codelion7:
I'm doing unit tests on it now. I'll have it done within the next 2 days.

PR #629, opened 2015-04-11
[l10n] Revision + New French v.1.6

fxbenard:
Here you are French 100% - It really needed it!
more to come later


pippinsplugins:
Thanks!

Issue #223, opened 2014-05-15
Email for pending affiliate


sumobi:
We don't currently have an email that is sent out to an affiliate when they have registered to become an affiliate, but their application is `pending`. Currently affiliates only get an email when their application has been approved.

I can add this into `class-register.php` where the admin notification is being sent

    if( affiliate_wp()->settings->get( 'registration_notifications' ) ) {
			affiliate_wp()->emails->notification( 'registration', array( 'affiliate_id' => $affiliate_id ) );
    // pending notification to affiliate here
		}

Should we make this an option in the plugin settings?

sumobi:
We should add a pending email to core

PR #638, opened 2015-04-12
Issue/223

sumobi:
Fixes https://github.com/affiliatewp/AffiliateWP/issues/223 and https://github.com/affiliatewp/AffiliateWP/issues/639


pippinsplugins:
The `std`s should be one translation string, not multiple. So instead of "Hi" . . . "{name} . . .". It should be "Hi {name} . . . .".

PR #643, opened 2015-04-12
Fix #608

### LINKED ISSUE #608: Update error validation
WordPress does not allow Uppercase letters in a username. When you try and do this it shows this message:

`Only lowercase letters (a-z) and numbers are allowed`
https://cloudup.com/c_oEuDjZ5vE

When you use uppercase letters in the username field of the affiliate registration form, it only shows a `Invalid username` message which has been noted as frustrating by a customer. We should mimmick the same message as WordPress when uppercase letters are used in the field.

sumobi:


sumobi:
I think this can go in a sooner release

PR #661, opened 2015-04-21
Add products tag to be used in email templates

ryanduff:


pippinsplugins:
I think we should show this one conditionally based on whether an integration that supports it is enabled, otherwise we will get a lot of tickets from people wondering why their products aren't showing.

ryanduff:
@pippinsplugins I was already considering something along those lines. Not sure what's best, but currently all registered tags show in admin for each email type. i.e. there are referral tags showing besides the registration email template. 

There should be some way to filter available tags by template type.

pippinsplugins:
I'd do it this way: https://github.com/affiliatewp/AffiliateWP/blob/master/includes/admin/settings/class-settings.php#L424

pippinsplugins:
@ryanduff Thoughts on the change proposal?

PR #664, opened 2015-04-21
Instantiate all objects on plugins_loaded

ryanduff:
For sanity... while we still can.

### DIFF
--- a/affiliate-wp.php
+++ b/affiliate-wp.php
@@ -178,23 +178,7 @@ final class Affiliate_WP {
 			self::$instance->setup_constants();
 			self::$instance->includes();
 
-			// Setup objects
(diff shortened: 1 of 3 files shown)


ryanduff:
I need to put more eyes on this. Overall things seem to work, but getting a fatal error on plugin activation.

Looks to be happening because of how register_activation_hook() runs early and the objects aren't instantiated til plugins_loaded.

ryanduff:
Added 2 more commits to handle install/upgrade and avoid fatal errors.

pippinsplugins:
Looks good to me. I'm going to test further and then merge if all is well.

pippinsplugins:
All Pro add-ons are working fine.

PR #675, opened 2015-04-28
register affiliatemeta table as early as possible

ryanduff:
Seems `init` is a bit late to register the affiliatemeta table with $wpdb, moving to `plugins_loaded` hook so that it's properly registered for (hopefully) all use cases. 

Had a function hooked into `affwp_post_update_affiliate` and when calling `affwp_update_affiliate_meta()` it was returning false because the table hadn't been registered for use.

### DIFF
--- a/includes/class-affiliate-meta-db.php
+++ b/includes/class-affiliate-meta-db.php
@@ -20,7 +20,7 @@ class Affiliate_WP_Affiliate_Meta_DB extends Affiliate_WP_DB {
 		$this->primary_key = 'meta_id';
 		$this->version     = '1.0';
 
-		add_action( 'plugins_loaded', array( $this, 'register_table' ) );
+		add_action( 'plugins_loaded', array( $this, 'register_table' ), 11 );
 
 	}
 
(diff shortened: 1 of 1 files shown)


pippinsplugins:
I think this would break then with the main AffiliateWP being loaded on `plugins_loaded`. Wouldn't it?

pippinsplugins:
If that's true, we should probably register this at the same time the main class is loaded (which would be on plugins_loaded).

ryanduff:
Tested against #664-- since `AffiliateWP` is hooked in at `plugins_loaded` default priority 10, this needs to run at 11 or later. Priority 10 will have already fired by the time the `add_action()` call happens so it's too late to add it.  

Tested with priority 11 and it's properly registering the table.

includes/class-affiliate-meta-db.php
```diff
@@ -20,7 +20,7 @@ public function __construct() {
 		$this->primary_key = 'meta_id';
 		$this->version     = '1.0';
 
-		add_action( 'init', array( $this, 'register_table' ) );
+		add_action( 'plugins_loaded', array( $this, 'register_table' ), 11 );
```
pippinsplugins: Shouldn't this be earlier to ensure it's registered before add-ons load?

ryanduff: You made me think about this for a second, but no. Reason being other plugins probably aren't (read: shouldn't) be doing any reading or saving of affiliate meta on instantiation but rather some later hook like `init`

pippinsplugins:
Confirmed that this was the cause / fix for difficulty in using the meta table here https://github.com/affiliatewp/affiliatewp-recurring-referrals/issues/7

PR #692, opened 2015-05-14
add support for query strings

### LINKED ISSUE #691: Support query strings in creatives and [affiliate_referral_url] shortcode
When a query string is used in the URL of a creative it converts the trailing slash to the HTML entity of `%2F`. This causes an error on apache servers.

EG this URL entered in the creative:

    http://site.com/?something=true

Becomes:

    <a href="http://site.com/?something=true%2F&ref=3">Link text</a>

Additionally, when `Pretty Affiliate URLs` is enabled in the settings, the URL becomes:

    http://site.com/?something=true/ref/3

This should be to 

    http://site.com/ref/3/?something=true

So overall, query strings can't be used in creatives or within the `url` parameter of the `[affiliate_referral_url]` shortcode. We'll need to rework how its currently done.

sumobi:
Fixes #691

### DIFF
--- a/includes/admin/creatives/actions.php
+++ b/includes/admin/creatives/actions.php
@@ -77,7 +77,7 @@ function affwp_process_update_creative( $data ) {
 	}
 
 	if ( ! current_user_can( 'manage_creatives' ) ) {
-		wp_die( __( 'You do not have permission to manage affiliates', 'affiliate-wp' ), __( 'Error', 'affiliate-wp' ), array( 'response' => 403 ) );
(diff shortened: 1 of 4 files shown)


pippinsplugins:
What's the purpose of the `query` key here?

sumobi:
@pippinsplugins if the base URL has a query string we want to remove it (in the case of pretty URLs) and then slam it on the end. parse_url() provides a query key in the array for getting the query string. This basically allow query string (even multiple) to be used in creatives, the referral URL shortcode etc.

PR #708, opened 2015-05-27
Add 4 actions to the referrals dashboard template

clickstudio:
These actions will allow plugins to extend the table columns and add other functions above and below the table. 

I am developing a plugin to convert affiliate commissions to 'Woocommerce Points & Rewards' points. I would like to add front end user actions to the mix.

### DIFF
--- a/templates/dashboard-tab-referrals.php
+++ b/templates/dashboard-tab-referrals.php
@@ -12,7 +12,7 @@
 		'status'       => array( 'paid', 'unpaid', 'rejected' )
 	) );
 	?>
-	<?php do_action( 'affwp_referrals_dashboard_before_referrals_table', affwp_get_affiliate_id() ); ?>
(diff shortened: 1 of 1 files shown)


templates/dashboard-tab-referrals.php
```diff
@@ -32,6 +34,7 @@
 						<td class="referral-description"><?php echo $referral->description; ?></td>
 						<td class="referral-status <?php echo $referral->status; ?>"><?php echo affwp_get_referral_status_label( $referral ); ?></td>
 						<td class="referral-date"><?php echo date_i18n( get_option( 'date_format' ), strtotime( $referral->date ) ); ?></td>
+						<?php do_action( 'affwp_referrals_dashboard_td', $referral->referral_id ); ?>
```
pippinsplugins: Could you change this to pass the whole referral object instead of just the ID?

clickstudio: Yeah no problem.

PR #713, opened 2015-05-31
Issue/711

### LINKED ISSUE #711: Add affiliate's username to Edit Affiliate screen
It would be nice to be able to see the affiliate's username on the `Edit Affiliate` screen, in the same fashion as the User ID:

screen shot 2015-05-28 at 12 27 37 pm

sumobi:
Fixes #711

### DIFF
--- a/includes/admin/affiliates/edit.php
+++ b/includes/admin/affiliates/edit.php
@@ -48,7 +48,7 @@ $email     = ! empty( $affiliate->payment_email ) ? $affiliate->payment_email : 
 				</th>
 
 				<td>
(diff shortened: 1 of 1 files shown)


pippinsplugins:
`$user_info->data->user_login` should just be `$user_info->user_login`. Why the `data` object?

sumobi:
@pippinsplugins user_login only exists in `data`

screen shot 2015-06-01 at 9 47 39 am

pippinsplugins:
Ah, pretty sure it uses a magic method to let you retrieve it with
`$user_data->user_login` any way.

sumobi:
Must be not so magic, that was the first thing I attempted but it returned null :)

sumobi:
Actually I think you're right, it works now. My magic must have been failing before.

pippinsplugins:
Ha ok then, merging as is :D

PR #714, opened 2015-05-31
Add word-wrap CSS rule to tables

### LINKED ISSUE #669: Long links flow outside the table in affiliate area
I think this sums it up fairly well

affiliatewp-1

sumobi:
Fixes #669 

To try and simulate the issue on themes that don't have suitable styling I first removed 2015's default CSS for word-wrap which resulted in this:

screen shot 2015-06-01 at 10 56 01 am

Then I added the new `word-wrap` CSS rule to our tables which gives us:

screen shot 2015-06-01 at 10 56 13 am

It would be better if themes had this CSS rule on `.entry-content`, ie the wrapping content div but because themes can easily change this class it would be better if we applied it specifically to our tables.

### DIFF
--- a/assets/css/forms.css
+++ b/assets/css/forms.css
@@ -186,3 +186,7 @@ p.affwp-error {
 	-ms-hyphens: none;
 	hyphens: none;
 }
+
+.affwp-table {
+	word-wrap: break-word;
+}


pippinsplugins:
Looks good to me.

PR #715, opened 2015-05-31
Remove the redirect from affwp_update_profile_settings()

clickstudio:
This redirect makes it impossible to process additional settings added by plugins.

I'm using the following action to add settings the user settings dashboard:  
add_action( 'affwp_affiliate_dashboard_before_submit', array( $this, 'front_end_settings_output' ), 10, 2 );

I can't find any way to process these settings due to the redirect.

### DIFF
--- a/includes/affiliate-functions.php
+++ b/includes/affiliate-functions.php
@@ -874,9 +874,7 @@ function affwp_update_profile_settings( $data = array() ) {
 
 	do_action( 'affwp_update_affiliate_profile_settings', $data );
 
-	if ( ! empty( $_POST['affwp_action'] ) ) {
-		wp_redirect( add_query_arg( 'affwp_notice', 'profile-updated' ) ); exit;
-	}
+	
 }
 
 /**


ryanduff:
I've been looking for a better approach to this as well thought I'm not sure this PR is the end-all, be-all fix. 

For a temporary workaround, hook in just before the redirect. Caveat being your extra settings are probably stripped from the array passed in `$data` so you'll need to use the `$_REQUEST

pippinsplugins:
Using the `affwp_update_affiliate_profile_settings` is the best method at this time. Removing the redirect is not an option, sorry. It changes behavior too much and removes the re-post protection.

clickstudio:
I have tried using:
add_action( 'affwp_update_affiliate_profile_settings', array( AffiliateWP_WooCommerce_Points::process_front_end_actions()));

The $data array does not get passed and the redirect occurs before this action can be registered by my plugin.

pippinsplugins:
Our action, affwp_update_affiliate_profile_settings, fires before the redirect.

Are you sure you've registered your add_action() early (or late) enough?

clickstudio:
I did find a small bug that was preventing the user meta from being updated due to the order of my actions however I did have to use $_REQUEST as $data does not pass through the new settings.

It seems to be an unfinished process but I have it working as expected. Sorry for wasting your time.

PR #725, opened 2015-06-16
Add option to credit last referrer

maurisrx:
When we use Clickbank, JVZoo, or Zaxaa, the last affiliate who refer visitors will get credited for sales. While in AffiliateWP, the first affiliate who refer visitor will get credited because it uses the first affiliate cookie referral variable.

Because I need this feature to make the last affiliate who refer visitors get credited, I add an option to choose whether user want to make the last or the first affiliate (while referral cookie still exists) who refers visitor that will get credited. The minified tracking.js hasn't been created yet because I don't know how you've done it. What tool do you use to minify js files?


includes/class-tracking.php
```diff
@@ -68,6 +68,7 @@ public function header_scripts() {
 		var AFFWP = AFFWP || {};
 		AFFWP.referral_var = '<?php echo $this->get_referral_var(); ?>';
 		AFFWP.expiration = <?php echo $this->get_expiration_time(); ?>;
+		AFFWP.credit_last_referrer = <?php echo ( ! empty( affiliate_wp()->settings->get( 'referral_credit_last' ) ) ) ? affiliate_wp()->settings->get( 'referral_credit_last' ) : 0 ?>;
```
fjarrett: @maurisrx This line is throwing a fatal error:

```
Fatal error: Can't use method return value in write context in includes/class-tracking.php on line 71
```

Please define `$referral_credit_last` before the `<script>` output in the `header_scripts()` method first.

Also, since the setting key is defined as `referral_credit_last` it would be better to use the same name in the `AFFWP` object too for consistency, instead of `AFFWP.credit_last_referrer`.

Something like this:

```php
public function header_scripts() {
	$referral_credit_last = affiliate_wp()->settings->get( 'referral_credit_last', 0 );
	?>
	<script type="text/javascript">
		var AFFWP = AFFWP || {};
		AFFWP.referral_var = '<?php echo $this->get_referral_var(); ?>';
		AFFWP.expiration = <?php echo $this->get_expiration_time(); ?>;
		AFFWP.referral_credit_last = <?php echo absint( $referral_credit_last ); ?>
	</script>
	<?php
}
```

pippinsplugins:
@maurisrx Sorry, I completely missed this PR. Overall looks good. If you could update for @fjarrett's suggestions, we can get this merged in.

maurisrx:
@fjarrett and @pippinsplugins, thank you for the suggestions. I've pushed a fix, please check it.

pippinsplugins:
@sumobi happy with this? If so, I'm ready to merge.

pippinsplugins:
Ready for 1.7 that is.

sumobi:
@pippinsplugins @maurisrx Works perfectly for me, great work.

Couple of minor points:

1. On the settings page lets reposition it underneath `Pretty Affiliate URLs`. Currently I think it breaks up the other settings that go together.

2. I think the description could be improved. How about instead of:

> Make the last referrer get credited instead of the first referrer whenever visitor makes a purchase.

We say:

> Credit the last affiliate who referred the customer

or 

> Credit the affiliate who referred the customer last

maurisrx:
@sumobi thanks! I've pushed another fix.

pippinsplugins:
@maurisrx Great great. We'll merge this in once 1.7 is closer to being ready.

sumobi:
:+1:

sumobi:
Thanks @maurisrx !

pippinsplugins:
I've manually merged this into `release/1.7`.

PR #750, opened 2015-06-30
Fix expected number_format() values on manual referral insert

### LINKED ISSUE #744: PHP Warning:  number_format() expects parameter 1 to be double
If you manually create a referral and leave the amount field blank:

    PHP Warning:  number_format() expects parameter 1 to be double, string given in /AffiliateWP/includes/misc-functions.php on line 120

fjarrett:
Fixes #744 

Simple sanitization of the first param value fixes the issue described with the `number_format()` function during manual referral creation.

This PR also implements sanitization on the filtered `$decimal` value, though not described in the issue it will prevent problems down the road as that param should always be a absolute integer.

**Aside:** Should manual referrals even be allowed when the amount is $0.00? Maybe that's a bit out of scope for the issue at hand, but something to think about anyway.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -117,7 +117,7 @@ function affwp_sanitize_amount( $amount ) {
 	}
 
 	$decimals = apply_filters( 'affwp_sanitize_amount_decimals', 2, $amount );
-	$amount   = number_format( $amount, $decimals, '.', '' );
+	$amount   = number_format( floatval( $amount ), absint( $decimals ), '.', '' );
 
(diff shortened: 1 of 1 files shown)


fjarrett:
@pippinsplugins I've probably issued this against the wrong branch. If so, let me know.

pippinsplugins:
Resubmit against `master`, otherwise looks good :)

fjarrett:
Done in https://github.com/affiliatewp/AffiliateWP/pull/752

PR #751, opened 2015-06-30
Issue/723 and Issue/765

### LINKED ISSUE #723: Entering a 1% commission rate changes rates to 100%
If you enter `2` in the `Referral Rate` field (general settings) all affiliate rates are set to `2%` as expected. If you enter `1` however, all affiliate rates are changed to `100%`. 

A workaround for the time being is to set the rate to `0.01`. Since 2 gives 2% I would expect 1 to give 1% also.

screen shot 2015-06-15 at 10 31 49 am

screen shot 2015-06-15 at 10 32 01 am

### LINKED ISSUE #765: A referral rate of 0 cannot be set unless you enter 0.00
Related: https://github.com/affiliatewp/AffiliateWP/issues/723

We've seen some customers who have wanted to disable the global commission rate entirely. They go ahead and enter `0` into the input field but the commission rate on the Affiliates -> Affiliates admin page still shows 20%. 

screen shot 2015-07-14 at 12 54 12 pm

screen shot 2015-07-14 at 12 54 25 pm

Entering `0.00` instead of `0` fixes this and the commission rate is correctly shown as `0%` on the affiliates screen. 

screen shot 2015-07-14 at 12 54 37 pm

screen shot 2015-07-14 at 12 54 48 pm

The commission rate input field however shows as `0` when saved, even after entering `0.00`.

We should allow `0` or `0.00` to be entered and have the commission rate set to 0%.

fjarrett:
Fixes #723
Fixes #765 

The `affwp_get_affiliate_rate()` function was improperly calculating rates that were 1% or less.

This PR is just a simple fix of logic to properly calculate any rate that is greater than `0.00`. So any numeric value entered in this field is now the literal percentage to expect.

 * Entering `1` means `1%`
 * Entering `0.01` means `0.01%`
 * Entering `100` means `100%`

I also went ahead and added HTML5 input validation, which is already [being done in the Settings](https://github.com/affiliatewp/AffiliateWP/blob/5e053dd6982f578034b27a37d09a74bc8b32b550/includes/admin/settings/class-settings.php#L239-L246), but wasn't being done when editing affiliates.

Also added the default rate as the placeholder on the input field when it is empty.

screen_shot_2015-06-30_at_12_00_18_pm


fjarrett:
@pippinsplugins @sumobi Again, just let me know if I've issued this against the wrong branch.

pippinsplugins:
Thanks @fjarrett!

While this works, I'm a bit concerned about users that have already configured it with `0.10` to mean `10%`. We permitted them to do that (perhaps stupidly) so we really can't assume that no one has. If someone has their rate set to `0.10` to mean 10%, they're going to have improper commission calculations once upgrading to 1.7.

I think your solution is 100% correct, we just have to figure out how to best handle this case so that no one has to go and update their rates.

Thoughts?

@sumobi Would you agree that there are almost certainly users using `0.10` to mean `10%`?

fjarrett:
The most complete fix would be a database scrub after the update.

sumobi:
@pippinsplugins Absolutely

How about showing a notice after the upgrade asking for confirmation?

"Thanks for upgrading! we noticed that your current rate is set at `0.1`. Due to recent changes in v1.7,  we've updated it to b `10`, as in `10 percent`. Please double-check the rate and adjust if incorrect." (click to dismiss)

fjarrett:
Between the time the upgrade happens and the notice is actually clicked there could already have been referrals happening with wrong rates.

Also if someone updated the plugin using an external tool, like ManageWP or something, they may never see the notice. This is an edge case, sure, but still a real possibility.

In general I think notices are good to do for DB alterations, but this one seems to be on the mission-critical side IMO.

We know the values affected by the bug, so it wouldn't be hard to isolate only those rows and update them automatically behind the scenes.

That's just what I would do :-)

fjarrett:
Re-reading your comment @sumobi I see you meant something else entirely. My bad.

sumobi:
All good. Yeah I meant exactly as you've described, update it automatically from `0.1` to `10` and show a notice explaining what we did. I doubt a customer would have wanted a 1 percent rate (0.1) so we should be safe to upgrade it to `10`. I still think we should show some kind of notice regardless, perhaps even link it to blog post.

fjarrett:
@sumobi Actually if a customer didn't know about the bug and entered `0.1` they would be wanting 1/10th of 1% - so an even _less likely_ case that would be desired.

I've added a DB migration script in https://github.com/affiliatewp/AffiliateWP/commit/068e0a9cb50b26418b9fa4a4dafbf533f0f2249f.

Please test it out and verify it works, then I can add some sort of notice.

sumobi:
@fjarrett I'm getting this (running a MS installation):

 WordPress database error Table 'affwp.wp_33_affiliate_wp_affiliates' doesn't exist for query SELECT affiliate_id, rate FROM wp_33_affiliate_wp_affiliates WHERE rate_type = 'percentage' AND rate > 0 AND rate <= 1 made by do_action('admin_init'), call_user_func_array, Affiliate_WP_Upgrades->init, Affiliate_WP_Upgrades->v17_upgrades

sumobi:
@fjarrett just spotted one issue so far:

Entering `1` for the affiliate commission rate gives affiliates a 100% commission rate. Entering `2` gives affiliates the expected 2% rate. Are you able to replicate that?

fjarrett:
@sumobi Ah I see that if the plugin is _network activated_ that there is one table for the whole network, otherwise blog-specific tables are used. I will fix this now.

fjarrett:
Hey @sumobi regarding the bug you found with `1` becoming `100`. Is the affiliate you are editing also happen to be your current logged user ID? What about editing other affiliates? I ask because I was able to reproduce that but only when editing myself. Very bizarre.

sumobi:
@fjarrett I was only testing with the global commission rate, and then looking at the commission rate column on the main affiliates page to see what the rate had changed to

fjarrett:
Alright @sumobi I have another round for you to test out. Now we are always treating referral rates as positive float values, and also allowing the value to be zero, or empty to use the site default rate.

Additionally, regarding the bug you discovered with `1` becoming `100` I figured out that's because the update routine is running continually in the background, which scrubs the database to fix any value less than or _equal to_ `1`.

To resolve, simply checkout this branch, load WP a few times, then go into your DB and update `affwp_version` to be `1.7`.

includes/admin/settings/class-settings.php
```diff
@@ -29,7 +29,9 @@ public function __construct() {
 	 * @return mixed
 	*/
 	public function get( $key, $default = false ) {
-		$value = ! empty( $this->options[ $key ] ) ? $this->options[ $key ] : $default;
+		$value = isset( $this->options[ $key ] ) ? $this->options[ $key ] : null;
+		$value = ( ! is_null( $value ) && '' !== $value ) ? $value : $default; // Allow 0 values
+
```
fjarrett: This is a scary line to be messing with, but it's necessary if we want to support `0` values. Needs thorough testing as this method is used everywhere throughout the whole plugin, not just referral rates.

pippinsplugins: Could it be modified in the `get_rate` functions instead?

pippinsplugins:
Could you bring this up to date with `release/1.7` so it merges cleanly?

includes/admin/settings/class-settings.php
fjarrett: @pippinsplugins [We are](https://github.com/affiliatewp/AffiliateWP/pull/751/files#diff-3538d6927ddef432e164ff73cf08dedeR193). But unfortunately we need this to be the behavior when calling [global settings values directly](https://github.com/affiliatewp/AffiliateWP/pull/751/files#diff-3538d6927ddef432e164ff73cf08dedeR196) from within `affwp_get_affiliate_rate()` as well as templates.

And I don't think we can abstract our own function just for this purpose either, because at the lowest level (this method) we are ignoring all empty values in favor of a default fallback.

Just to expand here a little. We specifically need to treat `null` and `string ''` as being truly "empty" and accept `int 0`, `string '0'`, and `float 0.00` as  acceptable values without reverting to the default.

pippinsplugins: Ah, good catch. Ok let's just test extensively.

sumobi:
@fjarrett @pippinsplugins will give this a thorough test tomorow

sumobi:
@fjarrett @pippinsplugins I hope you know that by "tomorrow" I actually meant 2 weeks.

So far this all checks out. I know you made it so if left blank it reverts to 20% @fjarrett, but shouldn't we just set it to 0% since that IS the default referral rate field. If I left it blank I personally would expect it to be nothing, ie 0. Either way I don't mind, they can enter 0 for 0% or leave blank and it will fallback to 20%.

sumobi:
Actually slight issue, the referral totals aren't being calculated correctly. For example: 

1% commission rate set
$100 purchase
referral = $0.01 (should have been $1.00)

5% commission rate set
$100 purchase
referral = $0.05 (should have been $5.00)

fjarrett:
@sumobi in regards to your [earlier comment](https://github.com/affiliatewp/AffiliateWP/pull/751#issuecomment-125505995):

1. Leaving an _affiliate's_ referral rate blank will fallback to the Referral Rate defined in the global settings.
2. Leaving the Referral Rate global setting blank will default to 20%, which is a [hardcoded](https://github.com/affiliatewp/AffiliateWP/blob/master/includes/admin/settings/class-settings.php#L245) default in the settings array.

fjarrett:
@pippinsplugins @sumobi OK I think I've stumbled upon a little contradiction in logic. There is a param called `$custom_rate` in the `affwp_get_affiliate_rate()` function that says:

 > `$custom_rate`  `string`  A custom rate that overrides site/affiliate settings

However inside the `affwp_calc_referral_amount()` function there is an inline comment just above the `affwp_get_affiliate_rate()` call that says:

 > If the affiliate has a custom rate set, use it. If no rate is specified, use the fallback.

What is the purpose of this 3rd `$custom_rate` param?

And probably more specifically, what is the priority/fallback order for rates?

pippinsplugins:
@fjarrett The custom rate is for passing in a product-specific rate

fjarrett:
@pippinsplugins So to confirm, is this the order of priority from highest to lowest?

1. Product-specific referral rate
2. Affiliate-specific referral rate
3. Global setting referral rate
4. Default global setting referral rate (hardcoded to 20%)

pippinsplugins:
I'm going to have to check with @sumobi for 1 and 2. We've gone back and forth on which should take precedence.

includes/admin/settings/class-settings.php
fjarrett: @pippinsplugins Alright remember yesterday when I couldn't get cookies to work? This change was causing that. Not good. So I came up with what I think is a much less-scary way to allow zero values, by explicitly whitelisting them by key name in a filterable array. See https://github.com/fjarrett/AffiliateWP/commit/3e122bd1b8e1dffea8c2fa2a287670fb942f9991

fjarrett: Whoops, fixed a hole in that logic in https://github.com/affiliatewp/AffiliateWP/commit/a16dc94c4faf3e6bc0127758688a0934c2f0e92d

sumobi:
@fjarrett @pippinsplugins 

Currently it's

1. Affiliate specific referral rate
2. Product-specific referral rate
3. Referral rate set in Affiliates -> Settings -> General

I still think a per-affiliate rate (#1) should override a product rate (#2). There's also other factors such as tiered affiliate rates and the affiliate product rates add-on (https://affiliatewp.com/addons/affiliate-product-rates/) which would actually appear higher than #1.

Started making a flow chart a while back that I intended to add to the docs, not finished:

screen shot 2015-07-30 at 9 03 53 am

fjarrett:
@sumobi Thanks for the diagram. You mention that if the tiered add-on is installed it should be given priority _over_ the affiliate rate, however in the diagram that is not the case, right?

sumobi:
@fjarrett sorry i meant the last part of my sentence about the affiliate product rates

fjarrett:
@sumobi Oh OK, so there are two different add-ons... got it.

sumobi:
Just to be even more confusing, yep!

includes/referral-functions.php
```diff
@@ -228,4 +231,60 @@ function affwp_count_referrals( $affiliate_id = 0, $status = array(), $date = ar
...
+ * 1.019 => (float) 1.02
+ * 1.1   => (string) 1.10
+ * 1.10  => (string) 1.10
+ * 1.199 => (string) 1.20
+ *
+ * @param  mixed  $val
+ * @param  int    $precision  Number of required decimal places (optional)
+ * @return mixed              Returns an int, float or string on success, null when empty
+ */
+function affwp_abs_number_round( $val, $precision = 2 ) {
```
fjarrett: I'm quite proud of this one...took forever to get it right :sunglasses:

fjarrett:
@sumobi Alright, this PR is ready for testing again. I've fixed the fallback order and the incorrect commission amounts you reported in an earlier comment.

includes/referral-functions.php
pippinsplugins: :D

sumobi: :beers:

sumobi:
@fjarrett awesome, will give it a thrash

sumobi:
@fjarrett I seem to be getting the same results as last time. I definitely have all your latest changes.

EDD Download = $100
global referral rate = 0.10%
referral amount = $0.00
expected amount = $0.10

EDD Download = $100
global referral rate = 1%
referral amount = $0.01
expected amount = $1.00

EDD Download = $100
global referral rate = 5%
referral amount = $0.05
expected amount = $5.00

Can someone give one of those a quick test and see if you get the same as me?

fjarrett:
OK @sumobi I think I've fixed it. There was a logic problem in the integrations base class that was using product rates over affiliate rates.

fjarrett:
I think this line is a little misleading. It makes it sound like product rates, in fact, take priority over individual affiliate rates...

screen_shot_2015-07-31_at_10_35_37_am

sumobi:
Looks good now @fjarrett, I'm getting the expected results - will continue testing.

@pippinsplugins @ramiabraham Could you guys give this a test also? The more eyes on this the better.

pippinsplugins:
Will do.

ramiabraham:
Functions as expected for me (nice work). I'd like to do a comprehensive doc on this so customers have it available, prior to this hitting a release branch. It'd also be prudent to ask the MLM plugin dev to confirm any potential issues that may be introduced from affwp_abs_number_round. (The original MLM plugin does not conflict, but there may be significant refactoring planned).

pippinsplugins:
@codelion7 Do you think you could test this with MLM?

codelion7:
I do agree that this should not conflict with MLM, but I'll definitely test this out to be sure.

pippinsplugins:
Opened #826 so that I can make changes.

PR #753, opened 2015-07-01
Issue/665

### LINKED ISSUE #665: Improved user searching
We should improve our user search that is shown when adding new affiliates, adding discount codes etc.

Currently when searching for a user via the affiliate discount fields, it will show all users, which means you can select the username of a pending affiliate.

Any other suggestions to improve the user search are welcome.

fjarrett:
Resolves #665 

This PR refactors the admin user search AJAX to be able to distinguish a specific set of affiliate users that should be included/excluded in query matches using a `data-affwp-status` input attribute.

**Some attribute value examples**

`none` - Only search users that **are not** affiliates (who have no affiliate status).
`any` - Only search users that are affiliates, with any status.
`active` - Only search users that are active affiliates.
`pending` - Only search users that are pending affiliates.

The only reserved values here are `none` and `any`, any custom status added in the future would be automatically supported.

Of course you can just omit this data attribute entirely and **all users** will be searched, affiliates and non-affiliates.

**Performance improvements**

I've also added a simple 500ms delay before AJAX requests are sent, rather than sending them immediately on the `keyup` event. This is a significantly more efficient way of achieving the same result.

### DIFF
--- a/includes/admin/ajax-actions.php
+++ b/includes/admin/ajax-actions.php
@@ -3,33 +3,81 @@
 // retrieves a list of users via live search
 function affwp_search_users() {
 
-	if( empty( $_POST['user_name'] ) ) {
+	if ( empty( $_POST['search'] ) ) {
 		die( '-1' );
 	}
 
-	if( ! current_user_can( 'manage_affiliates' ) ) {
+	if ( ! current_user_can( 'manage_affiliates' ) ) {
 		die( '-1' );
 	}
 
-	$search_query = htmlentities2( trim( $_POST['user_name'] ) );
+	$search_query = htmlentities2( trim( $_POST['search'] ) );
 
 	do_action( 'affwp_pre_search_users', $search_query );
 
-	$found_users = get_users( array(
-			'number' => 9999,
-			'search' => $search_query . '*'
-		)
+	$args = array();
+
+	if ( isset( $_POST['status'] ) ) {
+		$status = mb_strtolower( htmlentities2( trim( $_POST['status'] ) ) );
+
+		switch ( $status ) {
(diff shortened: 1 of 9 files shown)


fjarrett:
@sumobi There was a random error in [this Travis job](https://travis-ci.org/affiliatewp/AffiliateWP/jobs/69166292), it just needs to be restarted.

pippinsplugins:
Super awesome, can't wait to test this out!

includes/admin/ajax-actions.php
```diff
@@ -3,37 +3,85 @@
...
+						'status' => $status,
+					)
+				);
+				$args = array( 'include' => array_map( 'absint', wp_list_pluck( $affiliates, 'user_id' ) ) );
+		}
+	}
+
+	$found_users = array_filter(
+		get_users( $args ),
+		function ( $user ) {
```
fjarrett: Whoops, this requires PHP 5.3+, not sure what the compatibility rules are with AffiliateWP?

fjarrett: Based on our Slack convo yesterday, it sounds like we can't do it this way.

I have a feeling this entire issue will be replaced by #754 anyway which I plan to start working on tomorrow and will incorporate some of the same principles/functionality introduced in this PR.

@pippinsplugins So maybe hold off on looking deeper into this one for now.

pippinsplugins: Let's hold off for now.

pippinsplugins:
After playing with this, I really, really like this and am going to go ahead and merge it.

We will bump the minimum version to 5.3.

PR #755, opened 2015-07-01
Issue/741

### LINKED ISSUE #741: Remove pagination markup when not needed
When there's no pagination it still shows the markup in the HTML:

`<div class="affwp-pagination"></div>`

We should remove the HTML when not needed

fjarrett:
Fixes #741 

This PR will only show pagination in templates when there is more than 1 page worth of items found.


fjarrett:
@sumobi More random errors coming [from Travis](https://travis-ci.org/affiliatewp/AffiliateWP/builds/69166799). I think the build just needs to be restarted here too.

fjarrett:
@pippinsplugins OK this is also ready for code review.

PR #762, opened 2015-07-13
Issue/745

### LINKED ISSUE #745: Incorrect value in URL generator when using affiliate area shortcodes add-on
It was something we never planned for but apparently over 100 people like to split the affiliate area up using the shortcodes from https://wordpress.org/plugins/affiliatewp-affiliate-area-shortcodes/

The only issue that has been reported so far is that when the shortcode to display the "Profile Settings" is placed above the shortcode to output the "Referral URL Generator", the "Generate URL" button grabs the wrong value of the hidden field with an ID of `affwp-affiliate-id`. This is because we have 2 hidden fields with the same ID and it's grabbing the one from the profile settings.

line that grabs the wrong value:
https://github.com/affiliatewp/AffiliateWP/blob/master/assets/js/frontend.js#L7

We could make a small change to fix this:

`affId = $( this ).find( '#affwp-affiliate-id' ).val(),`

Or maybe we should simply rename the IDs on the inputs so they are never the same (HTML spec would agree here).

fjarrett:
Fixes #745

Now using classes for hidden input targeting within the proper scope of their wrapper.

And it turns out that the hidden input for Profile Settings doesn't need an ID at all. Only the `name` attribute is necessary for targeting the form server-side.

### DIFF
--- a/assets/js/frontend.js
+++ b/assets/js/frontend.js
@@ -3,8 +3,8 @@ jQuery(document).ready( function($) {
 	$( '#affwp-generate-ref-url' ).submit( function() {
 
 		var url                 = $( this ).find( '#affwp-url' ).val(),
-		    refVar              = $( this ).find( '.affwp-referral-var' ).val(),
-		    affId               = $( this ).find( '.affwp-affiliate-id' ).val(),
+		    refVar              = $( this ).find( 'input[type="hidden"].affwp-referral-var' ).val(),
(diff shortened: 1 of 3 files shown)


pippinsplugins:
Looks good. @sumobi could you test these changes?

sumobi:
@pippinsplugins @fjarrett Tested, looks good to me

Issue #508, opened 2015-01-15
Better anti-spam - reCaptcha 2.0 support


pippinsplugins:
We had a customer today report that the honey pot field on the registration form is getting bypassed by pesky spammers. We should add an option that lets admins choose whether to add an additional anti-spam type.

sumobi:
:+1:

pippinsplugins:
I vote we add support for the new reCaptcha: https://www.google.com/recaptcha/intro/index.html

sumobi:
@fjarrett @pippinsplugins works great for me

PR #763, opened 2015-07-13
Issue/508

fjarrett:
Implements #508 

Adds reCAPTCHA settings to the Misc settings tab:

screen_shot_2015-07-13_at_2_46_59_pm

When all three options are satisfied, Google reCAPTCHA will be shown during registration:

screen shot 2015-07-13 at 2 46 43 pm

If the reCAPTCHA is not verified, the user will receive an error:

screen shot 2015-07-13 at 2 51 35 pm

### DIFF
--- a/includes/admin/settings/class-settings.php
+++ b/includes/admin/settings/class-settings.php
@@ -415,7 +415,7 @@ class Affiliate_WP_Settings {
 						'desc' => __( 'Should affiliates be able to register accounts for themselves?', 'affiliate-wp' ),
 						'type' => 'checkbox'
 					),
-					'enable_recaptcha' => array(
+					'recaptcha_enabled' => array(
 						'name' => __( 'Enable reCAPTCHA', 'affiliate-wp' ),
 						'desc' => __( 'Would you like to prevent bots from registering affiliate accounts using Google reCAPTCHA?', 'affiliate-wp' ),
 						'type' => 'checkbox'
--- a/includes/admin/settings/class-settings.php
+++ b/includes/admin/settings/class-settings.php
@@ -138,7 +138,7 @@ class Affiliate_WP_Settings {
 				}
 			}
 		}
-		
+
 		// Loop through each setting being saved and pass it through a sanitization filter
 		foreach ( $input as $key => $value ) {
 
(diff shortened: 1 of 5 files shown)


fjarrett:
@pippinsplugins @sumobi Here is a working implementation of the feature, but any ideas you might have for improving it are welcomed.

includes/admin/settings/class-settings.php
```diff
@@ -138,7 +138,7 @@ function sanitize_settings( $input = array() ) {
 				}
 			}
 		}
-		
+
```
fjarrett: Sorry about all the whitespace lines. I have my IDE configured to automatically remove trailing whitespace and there is a lot of it throughout the AffiliateWP code. In general, I think this is a best practice and would suggest other devs do the same, if possible :smiley_cat: 

For a cleaner view of this PR ignoring whitespace: https://github.com/affiliatewp/AffiliateWP/pull/763/files?w=1

pippinsplugins: Removing white space is definitely best. I tend to forget about it at times :)

pippinsplugins:
Two things I see initially:

1. Let's add a filter to affwp_is_recaptcha_enabled() to ensure others can disable it programatically.

2. An error should be thrown if `affwp_is_recaptcha_enabled()` is `true` and the reCaptcha post data is not_ set

fjarrett:
@pippinsplugins Thanks! OK I've made the changes you suggested. Also moved the Google API verification logic into a private method.

includes/class-register.php
```diff
@@ -98,6 +98,14 @@ public function process_registration( $data ) {
 			$this->add_error( 'empty_tos', __( 'Please agree to our terms of use', 'affiliate-wp' ) );
 		}
 
+		if (
+			affwp_is_recaptcha_enabled()
+			&&
+			( empty( $data['g-recaptcha-response'] ) || empty( $data['g-recaptcha-remoteip'] ) || ! $this->recaptcha_response_is_valid( $data['g-recaptcha-response'], $data['g-recaptcha-remoteip'] ) )
```
pippinsplugins: To help clean this up and make it easier to read, I think the `empty` checks for the post data should be included in the  `recaptcha_response_is_valid` method.

templates/register.php
```diff
@@ -98,6 +98,14 @@
 			</p>
 		<?php endif; ?>
 
+		<?php if ( affwp_is_recaptcha_enabled() ) : ?>
+			<div class="g-recaptcha" data-sitekey="<?php echo esc_attr( affiliate_wp()->settings->get( 'recaptcha_site_key' ) ); ?>"></div>
+
+			<p>
+				<input type="hidden" name="g-recaptcha-remoteip" value=<?php echo esc_attr( $_SERVER['REMOTE_ADDR'] ); ?> />
```
pippinsplugins: `REMOTE_ADDR` isn't always set. This should use `affwp_get_ip()` instead.

fjarrett: @pippinsplugins Great point. [Is this](https://github.com/affiliatewp/AffiliateWP/blob/6f8557b61a15cd0d400489bf74114f7709588ee7/includes/class-tracking.php#L553) the one? Maybe it should become a global helper function instead since it's not really specific to tracking?

includes/class-register.php
fjarrett: @pippinsplugins Excellent idea. Done in https://github.com/affiliatewp/AffiliateWP/commit/4ec0092

templates/register.php
pippinsplugins: Yes that's it. You can use it globally like this: `affiliate_wp()->tracking-get_ip()`

fjarrett: Ah yes, the God object :wink:

fjarrett: Done in https://github.com/affiliatewp/AffiliateWP/commit/ad024ca

pippinsplugins:
Boom.

sumobi:
:bomb:

PR #770, opened 2015-07-14
Issue/727

### LINKED ISSUE #727: Email not sent on automatic affiliate registrations
`Automatically register new users as affiliates?` enabled
`Notify site admins of new affiliate registrations?` enabled 

If a user is automatically registered as an affiliate, the admin does not receive an email. The admin email function is currently only fired on the `affwp_register_user` hook, which only exists inside the `register_user` function.

Likewise, an affiliate will not receive a pending affiliate registration email if `Require that site admins approve affiliates before they can begin earning referrals?` is enabled. They will however receive the application accepted email if the setting is not active.

I think the admin email and the application pending email should be sent when users are automatically added as affiliates. Do you agree @pippinsplugins ?

fjarrett:
Resolves #727 

PR introduces a new hook to process emails for auto registered affiliates.

I modeled it after the `affwp_register_user` hook, so it just plugs in right beside it into the existing processing functions for approved, pending, and new affiliate emails.

### DIFF
--- a/includes/class-register.php
+++ b/includes/class-register.php
@@ -40,7 +40,7 @@ class Affiliate_WP_Register {
 		return apply_filters( 'affwp_register_form', ob_get_clean() );
 
 	}
-	
+
 	/**
 	 * Process registration form submission
 	 *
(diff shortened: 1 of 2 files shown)


fjarrett:

pippinsplugins:
Looks good to me! Merging!

PR #772, opened 2015-07-15
Issue/771

### LINKED ISSUE #771: Fatal error when Gravity Forms plugin is deactivated
Can we safely assume that users may accept referrals via Gravity Forms, and then later decide not to use the plugin anymore, or maybe not for a limited period of time?

Currently a fatal error will be thrown on the Referrals screen if:

1. Gravity Forms integration is enabled
2. Referrals exist that have a `gravityforms` context
3. Gravity Forms is deactivated

It's not a blocker because you can safely rid the error by unchecking the integration.

@sumobi Thoughts?

screen shot 2015-07-14 at 1 19 58 pm

fjarrett:
Fixes #771

Simple `class_exists()` checks for Gravity Forms classes resolves this issue.

Also added PHP docblocks with `@uses` to every method in this class so we can quickly know what Gravity Forms classes/methods are within them.

### DIFF
--- a/includes/integrations/class-gravityforms.php
+++ b/includes/integrations/class-gravityforms.php
@@ -2,6 +2,11 @@
 
 class Affiliate_WP_Gravity_Forms extends Affiliate_WP_Base {
 
+	/**
+	 * Register hooks for this integration
+	 *
+	 * @access public
+	 */
 	public function init() {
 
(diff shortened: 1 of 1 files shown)


fjarrett:

https://github.com/affiliatewp/AffiliateWP/pull/772/files?w=1

PR #774, opened 2015-07-15
Issue/677

### LINKED ISSUE #677: Action Hook Needed for Adding Custom Callbacks in Settings Class
When adding the Software Licensing system to AffiliateWP MLM, I couldn't find a way to add my own custom callback to the Settings class. My goal was to simply create a slightly modified version of the existing license callback, but without an action hook, I had no way to register the callback function.

I think an action hook would pave the way for more customized input fields to be added to the existing settings structure. What do you guys think?

PS. I did find a way to use the text callback, and modify it to behave like the license callback does, but it could be smoother.

fjarrett:
Resolves #677

Simple change that allows custom settings callbacks to be defined per option, and take priority over baked-in callback methods so they can be overridden when necessary.

So you can do this:

```php
function my_custom_settings( $settings ) {
	$settings['general']['custom'] = array(
		'name'     => 'Custom',
		'type'     => 'something',
		'callback' => 'my_something_callback'
	);

	return $settings;
}
add_filter( 'affwp_settings', 'my_custom_settings' );

function my_something_callback( $args ) {
	echo 'Do things';
}
```

### DIFF
--- a/includes/admin/settings/class-settings.php
+++ b/includes/admin/settings/class-settings.php
@@ -69,10 +69,12 @@ class Affiliate_WP_Settings {
 
 				$name = isset( $option['name'] ) ? $option['name'] : '';
 
+				$callback = ! empty( $option['callback'] ) ? $option['callback'] : array( $this, $option['type'] . '_callback' );
+
 				add_settings_field(
 					'affwp_settings[' . $key . ']',
 					$name,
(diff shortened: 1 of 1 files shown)


fjarrett:
How about this approach @codelion7 and @pippinsplugins? No hooks necessary, just allow for custom callbacks to be called on a per option

pippinsplugins:
@fjarrett I like that approach.

codelion7:
Very nice @fjarrett. That will work smoothly.

Issue #520, opened 2015-01-21
Add role selection to User Import


pippinsplugins:
It'd be great to have an option select the specific user roles that should be imported in the Users migration assistant, instead of just importing all users.

sumobi:
:+1:

PR #776, opened 2015-07-16
Issue/520

fjarrett:
Implements #520 

Now our users can select the roles they want to create affiliate accounts in the Migration area.

By default, all roles that contain users are automatically selected upon page load.

I've included a user count for each role for convenience, while also disabling those roles which contain no users, so they aren't even selectable.

screen shot 2015-07-15 at 7 19 03 pm

If a user unchecks all the roles, the button will not be clickable. Because that's just silly.

screen shot 2015-07-15 at 7 23 06 pm

I've also improved the end process a little by redirecting the user to the Affiliates screen with the "added successfully" message.

screen shot 2015-07-15 at 7 23 54 pm

### DIFF
--- a/includes/admin/tools/class-migrate-users.php
+++ b/includes/admin/tools/class-migrate-users.php
@@ -12,26 +12,32 @@ if ( ! defined( 'ABSPATH' ) ) exit;
 class Affiliate_WP_Migrate_Users extends Affiliate_WP_Migrate_Base {
 
 	/**
+	 * Migrate users belonging to these roles
+	 *
+	 * @var array
+	 */
+	public $roles = array();
+
+	/**
 	 * Process the migration routine
 	 *
-	 * @since 1.3
+	 * @since  1.3
+	 * @param  int    $step
+	 * @param  string $part
+	 * @param  array  $roles (added in 1.7)
 	 * @return void
 	 */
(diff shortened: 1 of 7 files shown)


fjarrett:

https://github.com/affiliatewp/AffiliateWP/pull/776/files?w=1

includes/admin/tools/class-migrate-users.php
```diff
@@ -42,64 +47,113 @@ public function process( $step = 1, $part = '' ) {
...
+			array(
+				'number'     => 100,
+				'offset'     => ( $step - 1 ) * 100,
+				'orderby'    => 'ID',
+				'order'      => 'ASC',
+				'meta_query' => array(
+					array(
+						'key'     => 'wp_capabilities',
+						'value'   => sprintf( '"(%s)"', implode( '|', $this->roles ) ),
+						'compare' => 'REGEXP'
```
pippinsplugins: I did NOT know you could do this!

pippinsplugins:
Works beautifully.

includes/admin/tools/class-migrate-users.php
fjarrett: Regex query on serialized data FTW 😎

PR #779, opened 2015-07-20
Issue/777

### LINKED ISSUE #777: Delete bulk action not working from dropdown below list table
If you try to delete one or more affiliates using the Bulk Actions dropdown menu below the List Table, the page just refreshes and nothing happens. The Bulk Actions dropdown _above_ the List table works fine.

screen shot 2015-07-16 at 7 45 38 am

I tried the same approach on Referrals and it worked for me. So this issue seems to be isolated to the Affiliates screen only.

fjarrett:
Fixes #777 

The delete affiliate bulk action now works from the bottom dropdown action.

However, I discovered another edge-case bug where:

1. The top bulk action was selected to Deactivate (or anything other than Delete)
2. The bottom bulk action was selected to Delete
3. Click on the "Apply" button next to the TOP bulk selection
4. The system would try to delete the selected affiliates

In general I think this would be somewhat rare, but I went ahead and accounted for it. So now if you want to bulk delete using either dropdown the other dropdown cannot be selecting any other action. WordPress core should really be accounting for this for us, but doesn't.

### DIFF
--- a/includes/admin/affiliates/affiliates.php
+++ b/includes/admin/affiliates/affiliates.php
@@ -297,7 +297,7 @@ class AffWP_Affiliates_Table extends WP_List_Table {
 	 * @return string Displays a checkbox
 	 */
 	function column_cb( $affiliate ) {
-		return '<input type="checkbox" name="affiliate_id[]" value="' . $affiliate->affiliate_id . '" />';
+		return '<input type="checkbox" name="affiliate_id[]" value="' . absint( $affiliate->affiliate_id ) . '" />';
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


fjarrett:

pippinsplugins:
@fjarrett Could we have just one `$action` variable setup please?

```php
if( isset( $_GET['action'] ) ) {
    
} elseif( isset( $_GET['action2'] ) ) {

}

if( ! empty( $action ) ) {
     // proceed
}

fjarrett:
@pippinsplugins OK sure, that will mean that the top bulk action dropdown will always take priority and doesn't address the edge case. Done in https://github.com/affiliatewp/AffiliateWP/commit/efc73bf

pippinsplugins:
@fjarrett in that case swap it around. I'd think bottom action should take priority, since people scroll from top to bottom. Agree?

fjarrett:
@pippinsplugins Oh yeah, that makes a lot of sense. Done in https://github.com/affiliatewp/AffiliateWP/commit/a770aac

pippinsplugins:
Sweet, will test soon.

pippinsplugins:
@fjarrett Could you update this so it merges cleanly?

fjarrett:
@pippinsplugins Done.

PR #784, opened 2015-07-21
Issue/783

### LINKED ISSUE #783: Incorrect total items found calculations when filtering visits
As I was working on #578 I discovered a problem with total items found calculations.

screen_shot_2015-07-20_at_5_45_21_pm

And upon looking into it further I found that the `count()` methods provided for Affiliates, Referrals, Creatives, and Visits are all actually duplicating query logic from their respective `get_*()` methods.

An easy way to consolidate the logic would be to add support for a second argument to every `get_*()` method.

#### Example

Take visits, for example. Setting a `$count` param to `true` would simply return the total number of results found, rather than the results themselves. This would be `false` by default.

```php
public function get_visits( $args = array(), $count = false ) {
```

So the entire `count()` method could just become:

```php
public function count( $args = array() ) {
	return $this->get_visits( $args, true );
}
```

This approach maintains all backwards compatibility for instances of:

```php
$count = affiliate_wp()->visits->count( array( 'some' => 'arg' ) );
```

fjarrett:
Fixes #783 

This PR has quite a lot of refactoring.

If it's too much, or if the approach is flawed in some way that I am not aware of, feel free to close with alternative recommendations.


pippinsplugins:
@fjarrett could you bring this up to date so it merges cleanly?

fjarrett:
@pippinsplugins Done.

Issue #785, opened 2015-07-21
Add per page screen option to Creatives


fjarrett:
Similar to #780 and #781

sumobi:
@fjarrett @pippinsplugins Tested and works great

PR #787, opened 2015-07-21
Issue/785

fjarrett:
Implements #785 

Adds a per page screen option to Creatives list table screen.


fjarrett:

Issue #786, opened 2015-07-21
Add per page screen option to Visits screen


fjarrett:
Similar to #780, #781 and #785

sumobi:
@fjarrett @pippinsplugins Works great for me

PR #788, opened 2015-07-21
Issue/786

fjarrett:
Implements #786 

Adds a per page screen option to Visits list table screen.


fjarrett:

Issue #624, opened 2015-04-09
Option to disable referrals for specific Ninja Forms


sumobi:
Just like Gravity forms https://github.com/affiliatewp/AffiliateWP/issues/609 we should add an option to disable referral creation on certain forms.

Issue #609, opened 2015-04-03
Disable referrals for certain forms in Gravity Forms


sumobi:
When Gravity Forms is set up, every form will generate a referral if an affiliate is being tracked. We should allow the user to disable certain forms from generating referrals

pippinsplugins:
Sounds good to me! Let's try to get it in 1.7

sumobi:
@pippinsplugins the PR above adds a checkbox to enable referrals for the form. 

![screen shot 2015-05-19 at 6 18 26 pm](https://cloud.githubusercontent.com/assets/52581/7696670/7a90d7da-fe53-11e4-878a-e15a42479293.png)

I think it makes more sense to have to enable referrals rather than disable referrals for every Gravity Form that is created. It also keeps the consistency with the other form elements which you have to "enable" to use.

Speaking of which, how do we handle backwards compatibility? Do we create an upgrade routine for v1.7 that grabs all current gravity forms, and enables the new checkbox option? This would prevent existing forms from suddenly not generating referrals.

pippinsplugins:
I'd be inclined to have a `disable` option simply for the backwards compatibility issue, but we could do an upgrade routine as well.

If we're going for consistency, let's write the description as `Enable referral creation for this form` so it matches the fields above it.

sumobi:
If we had a disable option then every time some one created a new form, they'd have to remember to disable referrals. I can't imagine all forms needing to generate referrals.

pippinsplugins:
True. Let's do an upgrade routine.

pippinsplugins:
@fjarrett mind testing this?

fjarrett:
@pippinsplugins Sure thing!

pippinsplugins:
This is now merged into `release/1.7`. Tagging as `needs testing` for one final testing round before closing it.

sumobi:
@pippinsplugins @fjarrett haven't been able to fault it so far

pippinsplugins:
@sumobi Did you test the upgrade routine as well?

PR #790, opened 2015-07-22
Issue/624

fjarrett:
Implements #624 

This PR adds a custom "Restriction" setting to Ninja Forms to enable/disable referral tracking.

screen_shot_2015-07-22_at_11_29_51_am

screen shot 2015-07-22 at 11 24 35 am

Just as we did in #609 (to maintain backwards compatibility) the default value for this setting is **checked**.

As such, I have also provided a DB upgrade script to set this setting for each form if it does not exist.

As an aside, there were also a lot of incorrect docblocks in the Ninja Forms integration class, so I fixed those up as well.

### DIFF
--- a/includes/integrations/class-ninja-forms.php
+++ b/includes/integrations/class-ninja-forms.php
@@ -5,9 +5,9 @@ class Affiliate_WP_Ninja_Forms extends Affiliate_WP_Base {
 	/**
 	 * Get thigns started
 	 *
-	 * @access  private
+	 * @access  public
 	 * @since   1.6
-	*/
+	 */
 	public function init() {
 
 		$this->context = 'ninja-forms';
@@ -24,9 +24,10 @@ class Affiliate_WP_Ninja_Forms extends Affiliate_WP_Base {
 	/**
 	 * Record referral on submission
 	 *
-	 * @access  private
+	 * @access  public
 	 * @since   1.6
-	*/
+	 * @param   int $sub_id
+	 */
 	public function add_referral( $sub_id ) {
 
 		global $ninja_forms_processing;
@@ -53,9 +54,10 @@ class Affiliate_WP_Ninja_Forms extends Affiliate_WP_Base {
 	/**
 	 * Restore a rejected referral when untrashing a submission
 	 *
-	 * @access  private
+	 * @access  public
 	 * @since   1.6
-	*/
+	 * @param   int $sub_id
+	 */
 	public function restore_referral( $sub_id = 0 ) {
 
 		if( ! affiliate_wp()->settings->get( 'revoke_on_refund' ) ) {
@@ -77,9 +79,10 @@ class Affiliate_WP_Ninja_Forms extends Affiliate_WP_Base {
 	/**
 	 * Revoke a referral when a submission is deleted or trashed
 	 *
-	 * @access  private
+	 * @access  public
 	 * @since   1.6
-	*/
+	 * @param   int $sub_id
+	 */
 	public function revoke_referral_on_delete( $sub_id = 0 ) {
 
 		if( ! affiliate_wp()->settings->get( 'revoke_on_refund' ) ) {
@@ -97,9 +100,12 @@ class Affiliate_WP_Ninja_Forms extends Affiliate_WP_Base {
 	/**
 	 * Build the reference URL
 	 *
-	 * @access  private
+	 * @access  public
 	 * @since   1.6
-	*/
+	 * @param   int    $reference
+	 * @param   object $referral
+	 * @return  string
+	 */
 	public function reference_link( $reference = 0, $referral ) {
 
 		if( empty( $referral->context ) || 'ninja-forms' != $referral->context ) {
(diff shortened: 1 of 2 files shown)


fjarrett:
@pippinsplugins Ready for code review.

includes/integrations/class-ninja-forms.php
```diff
@@ -16,40 +17,47 @@ public function init() {
...
 	public function add_referral( $sub_id ) {
 
 		global $ninja_forms_processing;
 
-		if( $this->was_referred() ) {
-
-			if ( $this->is_affiliate_email( $this->get_submitted_email() ) ) {
-				return; // Customers cannot refer themselves
-			}
+		if (
```
pippinsplugins: Can we break this up into multiple return statements please? Makes it much easier to read.

fjarrett: Done in https://github.com/affiliatewp/AffiliateWP/commit/d46ff4a7b9f23c15f0926202b661379bc66594a1

fjarrett:
@pippinsplugins Just to explain my thought process a little...I honestly don't think this setting really fits under "Display" or "Restrictions" but since Ninja Forms does not provide any hooks to add custom setting metabox sections, I had to pick one. Since this setting is a "allow" type setting, I decided on the Restrictions section.

pippinsplugins:
Restrictions is fine.

includes/integrations/class-ninja-forms.php
```diff
@@ -16,40 +17,50 @@ public function init() {
...
-				return; // Customers cannot refer themselves
-			}
+		if ( ! $ninja_forms_processing->get_form_setting( 'affwp_allow_referrals' ) ) {
+			return;
+		}
 
-			$description    = $ninja_forms_processing->get_form_setting( 'form_title' );
-			$total          = $this->get_total();
-			$referral_total = $this->calculate_referral_amount( $total, $sub_id, $ninja_forms_processing->get_form_ID() );
+		if ( ! $this->was_referred() ) {
```
pippinsplugins: Let's check if was referred first. It's a faster check. No need to perform a DB query if unnecessary.

fjarrett: Good call. I wondered the exact same thing. Done in https://github.com/affiliatewp/AffiliateWP/commit/1968470

pippinsplugins:
Looks good.

@sumobi can you test this further?

sumobi:
@pippinsplugins @fjarrett Works great for me. There's just one small issue I can see. We need to check for Ninja Forms or we'll get this:

> 23-Jul-2015 01:20:01 UTC] WordPress database error Table 'affwp.wp_46_nf_objects' doesn't exist for query SELECT id FROM wp_46_nf_objects WHERE type = 'form'; made by require_once('wp-admin/admin.php'), do_action('admin_init'), call_user_func_array, Affiliate_WP_Upgrades->init, Affiliate_WP_Upgrades->v17_upgrades, Affiliate_WP_Upgrades->v17_upgrade_nforms

sumobi:
As per https://github.com/affiliatewp/AffiliateWP/issues/609#issuecomment-103664942 I still feel this is something that should be turned on when creating a form, not enabled by default. For backwards compatibility it should be enabled for all existing forms. Does anyone agree? :)

fjarrett:
I don't have a strong opinion either way, but I tend to agree more with what you're describing @sumobi.

By turning it on by default we are making an assumption that the _majority_ of forms created will be for ordering, which seems unlikely to me. In actuality if we really wanted to make the default "enabled" then the checkbox should be for explicitly "disabling" it.

sumobi:
@fjarrett My thoughts exactly

pippinsplugins:
Seconded

fjarrett:
@sumobi Oh wow, thanks for the testing on multisite. It appears you can't "network activate" NF on multisite at all _and_ when you single-site activate it they are not creating blog-specific tables! Very interesting.

fjarrett:
@sumobi OK this PR should be good to go now for your multisite setup.

I also applied the same defaults we discussed in this thread to #609 issue via a new #792 PR.

pippinsplugins:
:+1:

sumobi:
@fjarrett I'm still seeing the notice above (https://github.com/affiliatewp/AffiliateWP/pull/790#issuecomment-123926657). I have NF installed on a sub-site, AffiliateWP is network activated.

fjarrett:
@sumobi Hmmm interesting. Can you check your DB for me and see if the NF tables have blog prefixes on them or not? e.g. `wp_42_nf_objects`

sumobi:
Prefixes, here's my Ninja Forms sub-site that I test on:

screen shot 2015-07-31 at 10 44 52 am

fjarrett:
@sumobi Thanks!

fjarrett:
OK @sumobi I've changed the upgrade script in https://github.com/affiliatewp/AffiliateWP/commit/fc42861de838ce4e59f8629d85012c22580c4b8a so it will now respect blog-specific tables. Please test again.

pippinsplugins:
:+1L

sumobi:
@fjarrett perfect, notice has gone away

PR #700, opened 2015-05-19
Issue/609

pippinsplugins:
#609

PR #792, opened 2015-07-23
Issue/609 (revised)

fjarrett:
This PR extends what was done in #700 for #609 in light of [this decision](https://github.com/affiliatewp/AffiliateWP/pull/790#issuecomment-123927628) made recently in #624.

I also went ahead and changed to using a `affwp_` prefix for the custom setting key rather than `gform_`. The latter makes it sound like it's a setting coming from Gravity Forms. I think it's more appropriate to use our prefix here. Let me know if you agree @pippinsplugins.

### DIFF
--- a/includes/admin/class-upgrades.php
+++ b/includes/admin/class-upgrades.php
@@ -134,11 +134,11 @@ class Affiliate_WP_Upgrades {
 
 					$meta = json_decode( $meta );
 
-					if ( isset( $meta->gform_allow_referrals ) ) {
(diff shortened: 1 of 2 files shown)


fjarrett:

pippinsplugins:
:+1:

fjarrett:
Oops I need to change the update script as well. Will do that now.

fjarrett:
Done

fjarrett:
@pippinsplugins Conflicts resolved

PR #795, opened 2015-07-23
Issue/791

### LINKED ISSUE #791: Affiliate emails being sent to the affiliate's payment email address
@pippinsplugins @fjarrett 

Emails such as the application pending and application accepted are being sent to the affiliate's payment email address. They should be sending to the affiliate's WP email address.

To replicate, simply sign up as an affiliate and enter both an email address and payment email address and you'll receive the emails at the payment email address. 

If you dump out an affiliate's email address onto the page you'll see it's grabbing the payment email:
https://github.com/affiliatewp/AffiliateWP/blob/master/includes/affiliate-functions.php#L336-L337

Since we need to get both the affiliate's email address in certain situations like sending emails, and we also need to grab the payment email we could pass in a new parameter to the function to determine which email we want. 

Ideally, and since this issue has cropped up before I think we should make a new `affwp_get_affiliate_payment_email()` function to keep things simpler, and update any code that needs the payment email address.

ticket: https://secure.helpscout.net/conversation/106671756/13317/

fjarrett:
Resolves #791 

 * Introduces a new `affwp_get_affilaite_payment_email()` function
 * Simplified the exiting `affwp_get_affilaite_email()` function
 * Removed the `get_affiliate_email()` method from the base integrations class that was not necessary
 * Simplified the `is_affilaite_email()` method in the base integrations class
 * The `$this->affiliate_id` property was being set inside other methods but it is already assigned by the `__construct()`...I'm not sure why they were done this way, so I removed them.

Both email getter functions now have a second parameter called `$default` which is optional and defaults to `false`. The idea was to use similar behavior to `get_option()` because there may be times when you really want the payment email, but if it's not available you're OK with using the user account email, or vice versa.

**Update:** I did the same for the `affwp_get_affiliate_login()` helper function too since it was using SQL queries unnecessarily.

### DIFF
--- a/includes/admin/tools/export/class-export-referrals-payout.php
+++ b/includes/admin/tools/export/class-export-referrals-payout.php
@@ -38,7 +38,6 @@ class Affiliate_WP_Referral_Payout_Export extends Affiliate_WP_Referral_Export {
 	public function csv_cols() {
 		$cols = array(
 			'email'         => __( 'Email', 'affiliate-wp' ),
-			'payment_email' => __( 'Payment Email', 'affiliate-wp' ),
 			'amount'        => __( 'Amount', 'affiliate-wp' ),
 			'currency'      => __( 'Currency', 'affiliate-wp' ),
 		);
@@ -90,10 +89,9 @@ class Affiliate_WP_Referral_Payout_Export extends Affiliate_WP_Referral_Export {
 				} else {
 
 					$data[ $referral->affiliate_id ] = array(
-						'email'         => affwp_get_affiliate_email( $referral->affiliate_id ),
-						'payment_email' => affwp_get_affiliate_payment_email( $referral->affiliate_id ),
-						'amount'        => $referral->amount,
-						'currency'      => ! empty( $referral->currency ) ? $referral->currency : affwp_get_currency()
+						'email'    => affwp_get_affiliate_payment_email( $referral->affiliate_id ),
+						'amount'   => $referral->amount,
+						'currency' => ! empty( $referral->currency ) ? $referral->currency : affwp_get_currency()
 					);
 
 					$affiliates[] = $referral->affiliate_id;
--- a/includes/admin/tools/export/class-export-referrals-payout.php
+++ b/includes/admin/tools/export/class-export-referrals-payout.php
@@ -37,9 +37,10 @@ class Affiliate_WP_Referral_Payout_Export extends Affiliate_WP_Referral_Export {
 	 */
 	public function csv_cols() {
 		$cols = array(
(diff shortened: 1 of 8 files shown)


includes/integrations/class-base.php
```diff
@@ -174,57 +174,37 @@ public function reject_referral( $reference = 0 ) {
...
-	}
-
-	/**
-	 * Retrieves the email address of the referring affiliate
-	 *
-	 * @access  public
-	 * @since   1.0
-	 * @return  string
-	 */
-	public function get_affiliate_email() {
```
pippinsplugins: We can't remove this because we don't know if any custom integrations are using it.

pippinsplugins:
I love you for adding unit tests :D

includes/integrations/class-base.php
fjarrett: @pippinsplugins OK thanks.

fjarrett:
:heart:

includes/integrations/class-base.php
fjarrett: @pippinsplugins Added back in https://github.com/affiliatewp/AffiliateWP/commit/ec6250178688e081ef20a6861ae6b4c671b4fd9b.

includes/admin/tools/export/class-export-referrals-payout.php
```diff
@@ -37,9 +37,10 @@ class Affiliate_WP_Referral_Payout_Export extends Affiliate_WP_Referral_Export {
 	 */
 	public function csv_cols() {
 		$cols = array(
-			'email'    => __( 'Email', 'affiliate-wp' ),
-			'amount'   => __( 'Amount', 'affiliate-wp' ),
-			'currency' => __( 'Currency', 'affiliate-wp' ),
+			'email'         => __( 'Email', 'affiliate-wp' ),
+			'payment_email' => __( 'Payment Email', 'affiliate-wp' ),
```
pippinsplugins: We actually have to remove this. PayPal doesn't support it (just applies to payout file)

includes/affiliate-functions.php
```diff
@@ -316,69 +316,74 @@ function affwp_get_affiliate_rate_types() {
...
 
+/**
+ * Retrieves the affiliate's payment email address
+ *
+ * @since  1.7
+ * @param  object|int $affiliate
+ * @param  mixed      $default (optional)
+ * @return mixed
+ */
+function affwp_get_affiliate_payment_email( $affiliate, $default = false ) {
```
pippinsplugins: This should retrieve the `email` if `payment_email` is not set. That way we can always just call `affwp_get_affiliate_payment_email()` when we need to get the payment email.

PR #800, opened 2015-07-28
Issue/799

### LINKED ISSUE #799: Gravity Forms: referral description is blank when no products are purchased
Related: https://github.com/affiliatewp/AffiliateWP/issues/609

When a referral is created using the Gravity Forms integration we pass the products into the referral's description. When we don't use products the description is blank: 

screen shot 2015-07-28 at 9 19 29 pm

See:
https://github.com/affiliatewp/AffiliateWP/blob/master/includes/integrations/class-gravityforms.php#L45

Instead of having a blank description I feel we should populate this with something useful like the form's title or description so you can easily see which form on the site generated the referral without having to click anywhere else. 

screen shot 2015-07-28 at 9 13 18 pm

The only issue I can see is if the admin named their form's title and description something private, and didn't necessarily want anyone to see it. If allow this, it will show up in the affiliate's referrals table on the front end. 

How about we add a new input field to the form settings, a place for the admin to add a label that will show as the referral's description, but only when products are not purchased?

@fjarrett @pippinsplugins

fjarrett:
Resolves #799 

The referral description will now default to the form title when no products exist in the form.

### DIFF
--- a/includes/integrations/class-gravityforms.php
+++ b/includes/integrations/class-gravityforms.php
@@ -44,7 +44,7 @@ class Affiliate_WP_Gravity_Forms extends Affiliate_WP_Base {
 
 			// Do some craziness to determine the price (this should be easy but is not)
 
-			$desc      = '';
+			$desc      = isset( $form['title'] ) ? $form['title'] : '';
 			$entry     = GFFormsModel::get_lead( $entry['id'] );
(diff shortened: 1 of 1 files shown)


fjarrett:
@pippinsplugins Ready for review. This is a fallback change, and is pretty minor.

PR #801, opened 2015-07-29
Additional WooCommerce hooks - Integrates WC Auto-Ship

clickstudio:
This will integrate AffiliateWP with the WC Auto-ship plugin.

I have tested it and it works well.

### DIFF
--- a/includes/integrations/class-woocommerce.php
+++ b/includes/integrations/class-woocommerce.php
@@ -24,6 +24,7 @@ class Affiliate_WP_WooCommerce extends Affiliate_WP_Base {
 		add_action( 'woocommerce_order_status_pending_to_processing', array( $this, 'add_pending_referral' ), 10 );
 
 		// There should be an option to choose which of these is used
+		add_action( 'woocommerce_payment_complete', array( $this, 'mark_referral_complete' ), 10 );
 		add_action( 'woocommerce_order_status_completed', array( $this, 'mark_referral_complete' ), 10 );
 		add_action( 'woocommerce_order_status_processing', array( $this, 'mark_referral_complete' ), 10 );
 		add_action( 'woocommerce_order_status_completed_to_refunded', array( $this, 'revoke_referral_on_refund' ), 10 );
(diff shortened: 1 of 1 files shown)


pippinsplugins:
Does AutoShip register those as custom hooks or are they simply (mostly) unused hooks in WooCommerce core?

clickstudio:
They are both hooks from the WooCommerce core.

After I posted this pull request I contacted Jon the developer to discuss the hooks a bit further.

I am waiting on his response but there may be the possibility of modifying WC Auto-Ship slightly to better suit if your preference is not to use these hooks.

clickstudio:
'woocommerce_payment_complete' is the hook that is triggered when payment is flagged as accepted by the payment gateway.

clickstudio:
The 'woocommerce_order_status_pending_to_processing' may not be ideal to create the pending referral as it is generally when the other hooks complete the referral.

Perhaps Jon could add another hook to WC Auto-Ship. 
Maybe even the existing one 'woocommerce_checkout_order_processed'.

I'll wait for Jon's thoughts on it.

jonrules:
That's correct, both woocommerce_payment_complete and woocommerce_order_status_pending_to_processing are core WooCommerce hooks.

I could add woocommerce_checkout_order_processed to WC Auto-Ship if needed. Is that still necessary after this pull?

clickstudio:
Thanks Jon,

No it wouldn't be required if the pull is accepted in its current form. 

It all depends on which hooks you are the AffiliateWP guys agree on.

pippinsplugins:
Can one of you explain the difference between `woocommerce_payment_complete` and `woocommerce_order_status_completed`? When does one fire but not the other?

clickstudio:
'woocommerce_payment_complete' is the hook that is triggered when payment is flagged as accepted by the payment gateway

'woocommerce_order_status_completed' would fire when the order status is marked as 'Completed'. This is a manual process by the site admin and could be days after the actual order has been paid for. I mark the orders as "Completed' once they are shipped.

clickstudio:
The WooCommerce 'update_status function builds dynamic actions for all possible order statues and order status changes.

See the screenshot

http://snag.gy/qFqdU.jpg

 If Jon hooks into the following just prior to the payment, we should be good to go. No need to modify AffiliateWP.

add_action( 'woocommerce_checkout_order_processed', array( $this, 'add_pending_referral' ), 10, 2 );

I'll test this.

clickstudio:
I just tested this and it works great!

See the screenshot Jon:

http://snag.gy/aGb3t.jpg

jonrules:
@clickstudio This looks good. I'll add the 'woocommerce_checkout_order_processed' hook to the 3.1.2 release in a couple days.

clickstudio:
Awesome. Thanks Jon.

PR #808, opened 2015-08-03
Fix for affiliate self referalls

douglas-srs:
Added a check to the filter 'affwp_is_customer_email_affiliate_email' so whenever it's false it will allow affiliates to receive commissions for their own purchases.

### DIFF
--- a/includes/class-tracking.php
+++ b/includes/class-tracking.php
@@ -536,7 +536,7 @@ class Affiliate_WP_Tracking {
 			$affiliate_id = $this->get_affiliate_id();
 		}
 
-		$is_self = is_user_logged_in() && get_current_user_id() == affiliate_wp()->affiliates->get_column( 'user_id', $affiliate_id );
+		$is_self = (is_user_logged_in() && get_current_user_id() == affiliate_wp()->affiliates->get_column( 'user_id', $affiliate_id )) && (apply_filters( 'affwp_is_customer_email_affiliate_email', $ret, $email, $this->affiliate_id ));
 
 		$active = 'active' === affwp_get_affiliate_status( $affiliate_id );
 


pippinsplugins:
I believe we already have a filter added for this in the upcoming 1.7. I will have to double check though.

douglas-srs:
Hi,

I'm not sure but if you are talking abou "add_filter(
'affwp_is_customer_email_affiliate_email', '__return_false' );", it only
works with the code I changed, thanks for rechecking!


----------------------------

Att,
*Douglas Soares*

pippinsplugins:
@sumobi could you give this a glance over?

douglas-srs:
Will it be possible to affiliates to receive commissions even without clicking their own referral links? That would help a lot because my affiliates will make purchases for their clients using their own account. If not possible that's ok, by fixing this bug I can manage the rest :)

sumobi:
@pippinsplugins Ok I see the issue, `affwp_is_customer_email_affiliate_email` only fires for a logged out user, since it's only included in the `is_affiliate_email()` function. What @douglas-srs is proposing is allowing this filter to run while logged in. Currently a logged-in affiliate cannot earn a commission on their own purchase when using that filter.

@douglas-srs `$ret`, `$email` and `$this->affiliate_id` aren't available to the filter within the `is_valid_affiliate()` function so you can't pass them as you've done unfortunately.

Since the `affwp_is_customer_email_affiliate_email` filter was only created for the `is_affiliate_email()` function I don't think it really suits. We should either create a separate filter for a logged in affiliate, or create 1 filter that both functions can use to cover both a logged out and logged in affiliate.

sumobi:
@douglas-srs 

> Will it be possible to affiliates to receive commissions even without clicking their own referral links? That would help a lot because my affiliates will make purchases for their clients using their own account. If not possible that's ok, by fixing this bug I can manage the rest :)

If affiliates are purchasing for their clients while logged in then the filter should fix this :)

pippinsplugins:
I agree with @sumobi that adding `affwp_is_customer_email_affiliate_email` to `is_valid_affiliate()` doesn't make sense.

The `affwp_is_customer_email_affiliate_email` filter runs for all customers, logged in and out. The email on the order info sent to AffWP by the eCommerce plugin is what is used, regardless of log in status.

If we make any change at all here, I'd propose we just add a new filter to the return of `is_valid_affiliate()`.

pippinsplugins:
Per comments above, I've introduced a new filter called `affwp_tracking_is_valid_affiliate` that will allow this to be done.

sumobi:
:+1:

PR #809, opened 2015-08-06
Add filter hook to make dev able to create custom field callback on settings page

maurisrx:
I am developing an add-on and found out that creating custom setting field callback is not quite easy and fast. So I add a new filter hook to filter function called for the callback. Here is an example use:

    add_filter( 'affwp_esp_license_callback', 'create_test_callback' );

    function create_test_callback( $callback ) {
        return 'test_callback';
    }

    function test_callback() {
        ?>
        <p>This is custom callback</p>
        <?php
    }

Please review. Thanks,
Mauris


pippinsplugins:
I don't see any reason for this.

To create a custom call back, all you need to do is supply the function name that renders the HTML field.

maurisrx:
Ah, okay...

I've managed to create custom callback now. Thanks!

-------- Original message --------

pippinsplugins:
@maurisrx glad to hear it!

PR #817, opened 2015-08-19
fixes for fragment identifier

### LINKED ISSUE #811: Incorrect referral URLs when using a # inside [affiliate_referral_url] shortcode and referral URL generator
When a hash (#) is used in the [affiliate_referral_url] shortcode it appends a slash (/) to the end of the URL which is incorrect:

    Shortcode:
    [affiliate_referral_url url="http://affwp.dev/2015/03/22/hello-world/#comment"]

    Shortcode output:
    http://affwp.dev/2015/03/22/hello-world/?ref=3#comment/ 

If you take away the trailing slash above it still does not correctly store the AffiliateWP cookies. Only without the hash (and everything after) does it work, eg:

    http://affwp.dev/2015/03/22/hello-world/?ref=3

As per the "Basics" section of https://en.wikipedia.org/wiki/Fragment_identifier we know that the above should have worked as the hash must always be at the end, after any query string arguments.

Using pretty URLs work correctly and the AffiliateWP cookies are stored in the browser.

    http://affwp.dev/2015/03/22/hello-world/ref/3/#comment 

And finally, there's an issue with using #'s in the Referral URL Generator. It should strip the fragment identifier out, and append it to the very end of the URL, after the query string. The correct URL should have been:

    http://affwp.dev/?ref=3#something

screen shot 2015-08-07 at 1 17 53 pm

sumobi:
- shift fragment identifier (#) to end of URL in referral URL generator
- prevent trailing slash if URL contains fragment identifier

Fixes #811


sumobi:
This has already been merged into 1.7

PR #860, opened 2015-09-27
Check for affiliate

### LINKED ISSUE #859: PHP Notice:  Undefined index: affiliate
PHP Notice:  Undefined index: affiliate in /wp-content/plugins/AffiliateWP/includes/class-tracking.php on line 264

Appears if you are logged in as an affiliate and then you try to use your own referral link.

sumobi:
+ formatting

Fixes #859

### DIFF
--- a/includes/class-tracking.php
+++ b/includes/class-tracking.php
@@ -263,7 +263,7 @@ class Affiliate_WP_Tracking {
 
 		$affiliate_id = isset( $_POST['affiliate'] ) ? absint( $_POST['affiliate'] ) : '';
 
-		if ( $this->is_valid_affiliate( $affiliate_id ) ) {
+		if ( ! empty( $affiliate_id ) && $this->is_valid_affiliate( $affiliate_id ) ) {
 
 			// Store the visit in the DB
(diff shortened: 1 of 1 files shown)


includes/class-tracking.php
```diff
@@ -261,9 +261,9 @@ public function prevent_canonical_redirect( $redirect_url, $requested_url ) {
 	 */
 	public function track_visit() {
 
-		$affiliate_id = absint( $_POST['affiliate'] );
+		$affiliate_id = isset( $_POST['affiliate'] ) ? absint( $_POST['affiliate'] ) : '';
```
pippinsplugins: Shoudn't we go ahead and bail here if no ID is supplied?

sumobi: Doesn't it bail anyway with the `die( '-2' )` since the affiliate ID isn't valid? Or are you thinking we should just `return` straight away if no affiliate ID is supplied? (makes sense).

pippinsplugins: I think we should bail immediately. No reason to run additional checks if none are needed.

sumobi: :+1:

Issue #876, opened 2015-10-06
URLs in Visits table incorrect with pretty permalinks enabled


ghost1227:
While working on #843, I discovered that with pretty permalinks enabled for affiliate URLs, the visits table displays landing pages inclusive of the ref variable as shown below:

![](http://cloud.section214.com/image/0I0O102F0v1z/Image%202015-10-05%20at%206.56.48%20PM.png)

Without pretty permalinks, the `?ref=x` query variable is removed before it is displayed. This is handled in the `includes/class-visits-db.php` file through the following line:

`$data['url'] = remove_query_arg( affiliate_wp()->tracking->get_referral_var(), $data['url'] );`

Removing the query arg this way doesn't work with pretty permalinks, so we'll have to find another way to handle it or add a check for pretty permalinks and strip the ref some other way.

ghost1227:
> get_query_var() only retrieves public query variables that are recognized by WP_Query. This means that if you create your own custom URLs with their own query variables, get_query_var() will not retrieve them

pippinsplugins:
We'll need to explode the URL if no query arg is found that matches the referral variable and then look to see if any of the URL pieces match the query var.

ghost1227:
That's the only place I see that matches that particular var. I should note that we're never actually defining that as a query var... which is why remove_query_arg isn't working.

ghost1227:
Though, since that's a configurable variable, I'm not sure how feasible defining it would be.

pippinsplugins:
It's intentionally not defined as a query var.

ghost1227:
So then if no query arg is found, explode the URL, check the pieces, and if one matches the query var, remove it and the following piece?

pippinsplugins:
Yep!

ghost1227:
@pippinsplugins dunno if you're going to like how i did it, but that PR fixes it.

ghost1227:
@pippinsplugins Updated unit tests

pippinsplugins:
@ghost1227 We'll include it in 1.8

PR #877, opened 2015-10-06
First pass at handling referral var removal in pretty permalinks

ghost1227:
Issue #876

### DIFF
--- a/includes/class-visits-db.php
+++ b/includes/class-visits-db.php
@@ -242,9 +242,26 @@ class Affiliate_WP_Visits_DB extends Affiliate_WP_DB {
 	public function add( $data = array() ) {
 
 		if( ! empty( $data['url'] ) ) {
+			$original_url = $data['url'];
+			$referral_var = affiliate_wp()->tracking->get_referral_var();
 
 			// Remove the referral var
-			$data['url'] = remove_query_arg( affiliate_wp()->tracking->get_referral_var(), $data['url'] );
+			$data['url'] = remove_query_arg( $referral_var, $data['url'] );
+
+			if( $original_url == $data['url'] ) {
+				$pieces = explode( '/', $data['url'] );
+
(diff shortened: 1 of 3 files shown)


includes/class-visits-db.php
```diff
@@ -242,9 +242,26 @@ public function count( $args = array() ) {
...
 
 		if( ! empty( $data['url'] ) ) {
+			$original_url = $data['url'];
+			$referral_var = affiliate_wp()->tracking->get_referral_var();
 
 			// Remove the referral var
-			$data['url'] = remove_query_arg( affiliate_wp()->tracking->get_referral_var(), $data['url'] );
+			$data['url'] = remove_query_arg( $referral_var, $data['url'] );
+
+			if( $original_url == $data['url'] ) {
```
pippinsplugins: Use a strict comparison here.

includes/class-visits-db.php
```diff
@@ -242,9 +242,26 @@ public function count( $args = array() ) {
...
+				$pieces = explode( '/', $data['url'] );
+
+				foreach( $pieces as $key => $piece ) {
+					if( $piece == $referral_var ) {
```
pippinsplugins: Instead of looping through, use a `strpos` check to see if the pattern is found. If `/$ref_var/` is found, run the URL through `substr()` to strip everything off the URL after the start of the ref var.

ghost1227: can we guarantee that the ref var is always last? that's why I didn't do it that way in the first place... thought it could potentially be another place in the string.

pippinsplugins: It could have query args after it, so let's only strip up to the next `/` (if there is one)

pippinsplugins: Sorry, next `/` or `?`

pippinsplugins:
Thanks @ghost1227 !

PR #878, opened 2015-10-06
Add unpaid referrals to Affiliates page

### LINKED ISSUE #798: Add "Unpaid Referrals" column to Affiliates page
A customer has requested an "Unpaid Referrals" column for the main Affiliates screen (Affiliates -> Affiliates) in addition to the "Paid Referrals" column that is already there.

ticket:
https://secure.helpscout.net/conversation/107896006/13629/

Here is their reasoning for this:

> Since there's a lag in paying many affiliate (e.g. we offer a 90 day money back guarantee, so we don't pay our affiliate for 90 days) We need to be able to track the unpaid referrals for all affiliate in addition to the paid referrals. It would be great if there was an additional column to quickly look at how all affiliate are doing without drilling down to each affiliate's page.

I think this would be nice, any objections?

ticket: https://secure.helpscout.net/conversation/107896006/13629/

ghost1227:
Fixes #798


includes/admin/affiliates/affiliates.php
```diff
@@ -222,7 +222,8 @@ public function get_columns() {
 			'affiliate_id' => __( 'ID', 'affiliate-wp' ),
 			'earnings'     => __( 'Earnings', 'affiliate-wp' ),
 			'rate'     	   => __( 'Rate', 'affiliate-wp' ),
-			'referrals'    => __( 'Paid Referrals', 'affiliate-wp' ),
+			'unpaid'       => __( 'Unpaid', 'affiliate-wp' ),
+			'referrals'    => __( 'Paid', 'affiliate-wp' ),
```
pippinsplugins: This should remain as `Paid Referrals` and `Unpaid Referrals`, otherwise the context doesn't really make sense.

ghost1227: I changed it because with the full words, it forces a line break on most monitors. We could adjust the width of the actions column to handle it, but it'd be a drop from 30% to 20% width... That ok?

pippinsplugins: Blarg. @sumobi thoughts?

sumobi: @pippinsplugins @ghost1227 I think we should keep the text the same but look at adjusting the column widths.

pippinsplugins: @ghost1227 Can you go ahead and revert the column names? We can then adjust the width like Andrew said.

ghost1227: @pippinsplugins @sumobi Done

PR #888, opened 2015-10-07
Change login for field IDs to prevent conflict with registration form

### LINKED ISSUE #884: Log in and registration form share same field IDs
When `allow registration` is enabled the affiliate area shows a login and registration form. When they are both shown, the field IDs of the `username` and `password` fields between the 2 forms are identical. If you click the `username` or `password` label on the login form, it jumps you up to the username or password field of the registration form. 

We should make sure the IDs are unique.

ghost1227:
Fixes #884

### DIFF
--- a/templates/login.php
+++ b/templates/login.php
@@ -1,4 +1,4 @@
-<?php 
+<?php
 global $affwp_login_redirect;
 affiliate_wp()->login->print_errors();
 ?>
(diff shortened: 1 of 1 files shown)


michaelbeil:
nice!

pippinsplugins:
Headsup: this might cause a few support tickets due to people that have written their own CSS. Unavoidable though.

michaelbeil:
solid.

PR #890, opened 2015-10-07
Add dismissable notice if no integrations are enabled

### LINKED ISSUE #722: Notice when no integrations are enabled
Slightly embarrassing; we recently had a site where, for some reason, our Exchange integration was disabled. This went unnoticed for a few days until our client noticed referrals weren't being processed.

It'd be great, if a notice would be displayed when there are no integrations enabled. This could be dismissible too, if a user is doing custom integrations. I'd be happy to submit a PR if you think this is something that could go in.

ghost1227:
Fixes #722

### DIFF
--- a/includes/admin/class-notices.php
+++ b/includes/admin/class-notices.php
@@ -5,10 +5,18 @@ class Affiliate_WP_Admin_Notices {
 	public function __construct() {
 
 		add_action( 'admin_notices', array( $this, 'show_notices' ) );
+		add_action( 'affwp_dismiss_notices', array( $this, 'dismiss_notices' ) );
 	}
 
 
 	public function show_notices() {
+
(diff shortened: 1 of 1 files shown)


includes/admin/class-notices.php
```diff
@@ -175,5 +183,18 @@ public function show_notices() {
 
 	}
 
+	/**
+	 * Dismiss admin notices when Dismiss links are clicked
+	 *
+	 * @since 1.7.5
+	 * @return void
+	 */
+	function dismiss_notices() {
```
pippinsplugins: This needs to have a nonce.

ghost1227: Adding. Ironically, I copied from EDD... which currently _doesn't_ have one.

ghost1227: @pippinsplugins fixed

PR #894, opened 2015-10-08
issue/893 – Hide delete users checkbox if only deleting self

### LINKED ISSUE #893: Deleting own affiliate account is unintuitive
If you select to delete your own affiliate account, it still displays the "Delete affiliate's user accounts as well?" checkbox. Of course, if you check it it bypasses the deletion preventing you from removing your own account, but the checkbox probably shouldn't show at all.

ghost1227:
Fixes #893


DrewAPicture:
Moved to #1218.

Issue #898, opened 2015-10-09
Referrals not marked as complete with PayPal in iThemes Exchange


pippinsplugins:
Using PayPal, referrals are created but they are never marked as complete.

sumobi:
Re-opening, it appears referrals are not being created at all in certain scenarios

pippinsplugins:
I have an idea why.

TimothyBJacobs:
The problem is that for PayPal Exchange uses transient transactions, so the session data might not be available when the transient transaction is converted to an actual transaction. The `it_exchange_generate_transaction_object` filter should be used to mark if the transaction was from an affiliate or not. Then that data can be used in the `it_exchange_add_transaction_success` hook. I can submit a PR if you'd like.

michaelbeil:
That would be great @TimothyBJacobs!

TimothyBJacobs:
Working on it now.

michaelbeil:
:+1:

pippinsplugins:
Thanks @TimothyBJacobs !

TimothyBJacobs:
No problem!

PR #908, opened 2015-10-12
Fix for PayPal and Exchange

TimothyBJacobs:
This is a fix for #898. Not 100% tested yet, though.


sumobi:
@TimothyBJacobs Thanks so much for this!

PR #930, opened 2015-10-24
Rewrite conditional so email can send

### LINKED ISSUE #929: Application accepted email never sends when require approval is not enabled
The application accepted email never sends to an affiliate when `Require approval` is not enabled. 

This is because the conditional is set to return if the `$old_status` is `pending`. When `require approval` is not enabled and an affiliate registers, both their `$status` and `$old_status` will be `active`, thus the email does not send.

sumobi:
Fixes #929

### DIFF
--- a/includes/emails/actions.php
+++ b/includes/emails/actions.php
@@ -86,7 +86,10 @@ function affwp_notify_on_approval( $affiliate_id = 0, $status = '', $old_status 
 		return;
 	}
 
-	if( 'active' != $status || 'pending' != $old_status ) {
+	// only send email if:
+	// affiliate approval is disabled ( $status = active, $old_status = active )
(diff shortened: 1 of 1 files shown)


includes/emails/actions.php
```diff
@@ -86,7 +86,10 @@ function affwp_notify_on_approval( $affiliate_id = 0, $status = '', $old_status
 		return;
 	}
 
-	if( 'active' != $status || 'pending' != $old_status ) {
+	// only send email if:
```
pippinsplugins: Can you redo this comment block with a style like this?
```
/*
 *
 *
/*
```

Also, the comment itself doesn't read very clearly to me.

sumobi: :+1:

PR #936, opened 2015-10-29
making sure we have a code before we check it

curtismchale:
In some instances this will try and check for a coupon code when there is no code set which ends up causing PMPro to throw errors at checkout.

### DIFF
--- a/includes/integrations/class-pmp.php
+++ b/includes/integrations/class-pmp.php
@@ -221,7 +221,7 @@ class Affiliate_WP_PMP extends Affiliate_WP_Base {
 
 		$affiliate_id = false;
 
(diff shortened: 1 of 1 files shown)


includes/integrations/class-pmp.php
```diff
@@ -221,7 +221,7 @@ public function get_coupon_affiliate_id( $coupon_code ) {
 
 		$affiliate_id = false;
 
-		if( pmpro_checkDiscountCode( $coupon_code ) ) {
+		if( isset( $coupon_code ) && ! empty( $coupon_code ) && pmpro_checkDiscountCode( $coupon_code ) ) {
```
pippinsplugins: This could be simplified into just `! empty( $coupon_code )` :)

Thanks!

curtismchale:
Simplified

pippinsplugins:
Thanks!

PR #969, opened 2015-11-17
First pass at simple registration options on user profile

pmgarman:
Sometimes (on large sites especially) adding affiliates from the default "add new" affiliates view doesn't quite work. Having an option right on the user profile could be handy and at the same time when viewing the user profile having some quick links to edit the affiliate or view ports may be useful as well.

This is a first pass at something I've already added for our own internal use. I've reworked it from our custom plugin to a part the core plugin. If it is valuable and worth merging in some feedback on how I could improve this to fit AffiliateWP better would be great and I'll update my PR. If it isn't really valuable in core... that's cool too and we'll just run with this as a separate plugin :) Just let me know!


pmgarman:
Some screenshots...

- User is not yet registered as an affiliate


- User is registered and an active affiliate


(active, inactive, pending, and rejected all get the same "actions" buttons currently. only unregistered affiliates get the register button)

michaelbeil:
This could be a worthy addition to core.

pippinsplugins:
I've manually merged this into `release/1.8`.

Thanks @pmgarman !

PR #979, opened 2015-11-19
calculate product referral rates

### LINKED ISSUE #978: Referrals generated from Zippy Courses ignore commission rates
Looks like referrals generated from Zippy Courses are created at the product price. It doesn't take into consideration the referral rates (either global or per-product).

https://github.com/affiliatewp/AffiliateWP/blob/master/includes/integrations/class-zippycourses.php#L169

sumobi:
Fixes #978

### DIFF
--- a/includes/integrations/class-zippycourses.php
+++ b/includes/integrations/class-zippycourses.php
@@ -166,8 +166,6 @@ class Affiliate_WP_ZippyCourses extends Affiliate_WP_Base {
                     return; // Customers cannot refer themselves
                 }
 
-                $total = $order->getTotal();
-
                 $product = $order->getProduct();
 
                 // calculate referral amount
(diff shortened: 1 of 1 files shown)


includes/integrations/class-zippycourses.php
```diff
@@ -166,12 +166,18 @@ public function add_pending_referral( Zippy_Event $event ) {
                     return; // Customers cannot refer themselves
                 }
 
-                $total          = $order->getTotal();
+                $total = $order->getTotal();
```
pippinsplugins: Looks like this isn't used anywhere.

sumobi: Correct, it is not. I'm pretty sure there's no "cart", so only 1 product can be purchased at any one time. Because of this, we could either use the product amount, or the total. I'll remove it.

PR #1017, opened 2015-12-15
Fix redirects

### LINKED ISSUE #1005: WishList Member blocks emails, redirects affiliate when specific option is enabled.
We don't have a full integration for Wishlist Member but there's a couple of issues caused by WLM when the  "Allow WishList Member To Handle Login Redirect" option is enabled from WishList Member -> Settings -> Miscellaneous. 

1. It redirects the affiliate to the site's homepage, instead of keeping them on the affiliate area

2. It blocks emails being sent to the admin and affiliate when an affiliate registers. This is most likely due to the redirect, our email actions probably don't have time to fire.

sumobi:
Fixes #1005

### DIFF
--- a/includes/plugin-compatibility.php
+++ b/includes/plugin-compatibility.php
@@ -2,7 +2,7 @@
 
 /**
  *  Prevents OptimizeMember from intefering with our ajax user search
- *  
+ *
  *  @since 1.6.2
  *  @return void
  */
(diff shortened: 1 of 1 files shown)


includes/plugin-compatibility.php
```diff
@@ -11,4 +11,26 @@ function affwp_optimize_member_user_query( $search_term = '' ) {
...
+/**
+ *  Fixes affiliate redirects when "Allow WishList Member To Handle Login Redirect"
+ *  and "Allow WishList Member To Handle Logout Redirect" are enabled in WishList Member
+ *
+ *  @since 1.7.13
+ *  @return boolean
+ */
+function affwp_wishlist_member_redirects( $return ) {
+
+    $user    = wp_get_current_user();
```
pippinsplugins: Will the user always be logged in here or should there be a check first?

sumobi: @pippinsplugins Yep they will be, here's the WLM function with the filter included: https://cloudup.com/cZ88j5xalZt

pippinsplugins: :+1:

PR #1023, opened 2015-12-16
Improved user search

### LINKED ISSUE #1021: affwp_search_users ajax callback chokes on database with large number of users.
This callback method retrieves ALL wp_users (excluding any already existing affiliates) and then runs it through `array_filter` to match the search query against it.  
However on databases with a large user base it chokes because of the sheer number of users being pulled (I ran into this while testing AffiliateWP for a possible implementation on EventEspresso.com).  

Is there a reason why the search match is being done this way vs doing a direct `LIKE '%$search_string%'` against the user table?

If desired I can do a pull request to convert to a direct query?

nerrad:
Use sql to search instead of grabbing (potentially) large number of users from db - fixes #1021

### DIFF
--- a/includes/admin/ajax-actions.php
+++ b/includes/admin/ajax-actions.php
@@ -48,25 +48,22 @@ function affwp_search_users() {
 		}
 	}
 
-	$found_users = array_filter(
-		get_users( $args ),
-		function ( $user ) {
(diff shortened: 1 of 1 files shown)


pippinsplugins:
Works great!

PR #1041, opened 2015-12-30
Allow hhvm to fail tests

michaelbeil:


DrewAPicture:
FWIW, we've been seeing a lot more failures on the PHP7 front than HHVM.

michaelbeil:
Agreed. This is from a while ago and must have been missed. Closing.

Issue #1054, opened 2016-01-14
Emails as usernames display incorrectly


sumobi:
When an affiliate uses an email address on the standard affiliate registration form it is shown incorrectly on the affiliate area: 

![screen shot 2016-01-15 at 10 10 19 am](https://cloud.githubusercontent.com/assets/52581/12337678/3e2d9fb4-bb70-11e5-9a33-b4557a66c44c.png)

sumobi:
:+1:

PR #1059, opened 2016-01-15
Decode affiliate URL so that emails are shown properly. #1054

pippinsplugins:
#1054


sumobi:
@pippinsplugins just needs the `Referral URL` field to show the URL correctly also

PR #1072, opened 2016-01-24
Improve affwp_integration_create_referral filter

### LINKED ISSUE #1071: Improve affwp_integration_create_referral filter
We should improve the `affwp_integration_create_referral` filter so it's more useful. Context is pretty useless, so let's provide the filter with affiliate ID and reference.

sumobi:
Fixes #1071


includes/integrations/class-base.php
```diff
@@ -68,7 +68,10 @@ public function was_referred() {
 	 */
 	public function insert_pending_referral( $amount = '', $reference = 0, $description = '', $products = array(), $data = array() ) {
 
-		if ( ! (bool) apply_filters( 'affwp_integration_create_referral', true, $this->context ) ) {
+		// get affiliate ID
+		$this->affiliate_id = isset( $data['affiliate_id'] ) ? $data['affiliate_id'] : $this->get_affiliate_id( $reference, $this->context );
+
+		if ( ! (bool) apply_filters( 'affwp_integration_create_referral', true, array( $this->affiliate_id, $amount, $reference, $description, $products, $data ) ) ) {
```
pippinsplugins: This will need to setup each key/value for the array, otherwise it won't be possible to know which value belongs to which key. Doing it like this would cause something like the following:

array( 0 => 4, 1 => 12, 2 => 2233, 3 => 'edd' )

sumobi: Good point

PR #1103, opened 2016-02-15
Modify return type of affwp_add_referral()

JustinSainton:
affwp_add_referral() becomes a bit more helpful as a wrapper function for `affiliate_wp()->referrals->add( $args )` if it matches the return value.

Now, we can use the returned value of the referral to actually do something with the referral that was just added.  Woot!

All the falsey/truthy checks in core (maybe 1?) will still work fine, as 0 evaluates to false, and any other positive integer evaluates to true.

### DIFF
--- a/includes/referral-functions.php
+++ b/includes/referral-functions.php
@@ -116,26 +116,26 @@ function affwp_set_referral_status( $referral, $new_status = '' ) {
  * Adds a new referral to the database
  *
  * @since 1.0
- * @return bool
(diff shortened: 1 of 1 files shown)


TimothyBJacobs:
Two other possibilities without altering the return type.

1. Add a 2nd parameter `&$referral_id = null` and set that to the new referral ID.
2. New function `affwp_add_referral_and_get_id()`

pippinsplugins:
I definitely like returning the same datatype.

PR #1109, opened 2016-02-18
issue/933 – Affiliate Dashboard: Make vist urls human friendly

### LINKED ISSUE #933: Simplify URL Visit Column
In the front-end, it'd be great if a more simple URL could be displayed. This would be by stripping the domain and the referral arg.

Before:
`http://www.mysite.com/product/product-name/ref/username`

After:
`/product/product-name/`

All of that extraneous info means that the column almost necessarily wraps, and is difficult to understand.

Thoughts?

TimothyBJacobs:
URLs are now shortened to remove unnecessary information, like the host
or non public query vars. The url is now a link as well.

Introduces `affwp_make_url_human_readable()`.

Fixes #933.


DrewAPicture:
Moved to #1217 to diff against `release/1.8`.

PR #1110, opened 2016-02-18
update Ninja Forms integration for compatibility with Ninja Forms THREE.

kjohnson:
Deprecated old methods with a Ninja Forms version check for backwards compatibility and added the new Ninja Forms action class for supporting Ninja Forms THREE (v3).

### DIFF
--- a/includes/integrations/class-ninja-forms-action-add-referral.php
+++ b/includes/integrations/class-ninja-forms-action-add-referral.php
@@ -1,0 +1,142 @@
+<?php if ( ! defined( 'ABSPATH' ) || ! class_exists( 'NF_Abstracts_Action' )) exit;
+
+/**
+ * Class Affiliate_WP_Ninja_Forms_ActionAddReferral
+ */
+final class Affiliate_WP_Ninja_Forms_ActionAddReferral extends NF_Abstracts_Action
+{
+    /**
+     * @var string
+     */
+    protected $_name  = 'affiliatewp_add_referral';
+
+    /**
+     * @var array
+     */
+    protected $_tags = array( 'affiliate', 'affiliatewp', 'referral' );
+
+    /**
+     * @var string
+     */
+    protected $_timing = 'late';
+
+    /**
+     * @var int
+     */
+    protected $_priority = '10';
+
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+        parent::__construct();
+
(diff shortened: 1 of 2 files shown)


pippinsplugins:
@kjohnson Will this action register automatically or does the site owner need to enable it (like MailChimp)?

kjohnson:
The `Add Referral` Action will be available if the Ninja Forms integration is enabled in the AffiliateWP settings. It will not be added to the form automatically, but it can be added. This is apposed to adding the form setting convention in 2.x.

screenshot 2016-02-18 10 14 53

ramiabraham:
In testing, I don't see any AffiliateWP fields being present, or any AffiliateWP functionality within NF3.

- Added code from this pull into https://github.com/AffiliateWP/AffiliateWP/tree/issue/1110
- Accounted for the corrected NF typo (https://github.com/wpninjas/ninja-forms/issues/1472)
- Results in db errors, root cause being a missing table `wp_nf3_field_meta` ( @kjohnson is this a known issue at all?)
- Using affwp at https://github.com/AffiliateWP/AffiliateWP/commit/06e059e330c17bdf5eabf7491c6e75b20dea44fd, NF:master at `70276324`

kjohnson:
@ramiabraham the integration adds a new Form Action type to Ninja Forms [see above screenshot]. This new actions have settings for mapping the total, as well as settings for referrer email address, etc.

Also, the database tables are created when upgrading from NF2 to NF3. This can also be done by enabling "developer mode" and running "mock data", see http://developer.ninjaforms.com/codex/mock-data/

ramiabraham:
Thanks @kjohnson. We were able to track the cause of referrals not being created to the `$reference` being a non-unique string (which is required in AffiliateWP).

I'd like to use the NF form submission ID as the unique string for referrals (instead of the datetime string placeholder in there now).

In NF 2.x, you could do something like:

`$ninja_forms_processing->get_form_setting( 'sub_id' )`

Do you know what the equivalent would be in NF3?

Looks like it's too early to query the object with the built-in NF3 methods, eg
`Ninja_Forms()->form()->sub($nf_sub_id )->get()`

(latest is https://github.com/AffiliateWP/AffiliateWP/commit/6e39ab51510788c898b8981ba7303aac4a706817)

ramiabraham:
_(Closing in favor of https://github.com/AffiliateWP/AffiliateWP/pull/1421/)._

PR #1143, opened 2016-03-13
issue/934 – New parameters for login and registration shortcodes

### LINKED ISSUE #934: Should we make the affiliate_login shortcode redirect to affiliate area by default?
A lot of customers open support tickets about seeing a white page when an affiliate logs in and they always ask how can an affiliate be redirected to the affiliate area instead. The solution is to add the redirect parameter and link it to the affiliate area. 

The reason we don't automatically redirect is in case the site owner wants to keep the affiliate on the same page and show some kind of content using the [affiliate_content] shortcode. I'm not sure if anyone actually does this. Should we leave as is, or redirect an affiliate by default to the affiliate area? Any other solution to reduce the number of support tickets?

sumobi:
Fixes #934


includes/class-shortcodes.php
```diff
@@ -91,11 +116,17 @@ public function affiliate_login( $atts, $content = null ) {
...
+		$atts = shortcode_atts(
+			array(
 				'redirect' => '',
-			), $atts, 'affiliate_registration' )
+			),
+			$atts,
+			'affiliate_registration'
 		);
 
+		$redirect = isset( $atts['redirect'] ) ? $atts['redirect'] : '';
```
pippinsplugins: Let's use `! empty()` here instead of `isset()`. Using `isset()` could return true even if the redirect value is empty.

Issue #1144, opened 2016-03-13
Add affiliate while creating new user


sumobi:
Wanted to add some affiliates and got sick of filling out the affiliate registration form, or having to create a new user first and then adding them as an affiliate. How do we feel about an `Add as Affiliate` option directly on the create new user page?

Multisite:

![screen shot 2016-03-13 at 8 56 05 pm](https://cloud.githubusercontent.com/assets/52581/13727623/3ce12998-e95e-11e5-944b-3973a149c589.png)

On multisite this is only going to be possible (as far as I can see) if the `Skip Confirmation Email` option is enabled. We could add a better description on the checkbox description, or only show the checkbox when the skip option is selected.
___

Single site:

![screen shot 2016-03-13 at 8 56 37 pm](https://cloud.githubusercontent.com/assets/52581/13727622/25191e6a-e95e-11e5-969e-71de82de5cdd.png)

wordmaid:
This is freakin' awesome. It would also be cool if you could then skip/disable the _Affiliate registration accepted_ email which would be sent to affiliates when their affiliate account is created.

pippinsplugins:
Love it!

sumobi:
On multisite I've added it to the top `Add Existing User` section also since it would be also nice to add a user from another sub-site and then instantly add them as an affiliate. 

On each section, the `Add as Affiliate` checkbox is only shown when the `Skip Confirmation Email` checkbox is selected.

![screen shot 2016-03-15 at 3 17 11 pm](https://cloud.githubusercontent.com/assets/52581/13765749/173a3324-eac1-11e5-84b7-9bd9d23e1ee9.png)

___

After clicking the `Skip Confirmation Email` checkbox:

![screen shot 2016-03-15 at 3 17 31 pm](https://cloud.githubusercontent.com/assets/52581/13765755/1feec318-eac1-11e5-85df-cd7f2498b290.png)

sumobi:
Tested on multisite and single site installations

michaelbeil:
All checks out well here http://beil.io/1apLR.

sumobi:
@pippinsplugins think we should add an additional checkbox as per https://github.com/AffiliateWP/AffiliateWP/issues/1144#issuecomment-195949798? 

This would be an option to disable the Affiliate Application Accepted email that is sent to the affiliate.

pippinsplugins:
Ah yes, let's do.

sumobi:
[...]

PR #1149, opened 2016-03-15
Add affiliate while creating new user

sumobi:
#1144


michaelbeil:
All added strings are translatable too :smile:.

Issue #1116, opened 2016-02-25
RCP: Undefined indexes when registering while logged in


pippinsplugins:
I'm seeing the following notices when purchasing a subscription in RCP while logged in:

Notice: Undefined index: user_name in /Users/pippin/Sites/edd/wp-content/plugins/affiliate-wp/includes/integrations/class-rcp.php on line 249

Notice: Undefined variable: user_id in /Users/pippin/Sites/edd/wp-content/plugins/affiliate-wp/includes/integrations/class-rcp.php on line 258

michaelbeil:
Oh noes.

sumobi:
This seems to be happening when you purchase a subscription while logged in, and with a discount code.

When a subscription is purchased it's firing the `increase_uses` function:
https://github.com/restrictcontentpro/restrict-content-pro/blob/master/includes/class-rcp-discounts.php#L220 

This in turn is firing the `update()` function:
https://github.com/restrictcontentpro/restrict-content-pro/blob/master/includes/class-rcp-discounts.php#L344

Inside the update function is the `rcp_edit_discount` action hook:
https://github.com/restrictcontentpro/restrict-content-pro/blob/master/includes/class-rcp-discounts.php#L392

AffiliateWP uses this `rcp_edit_discount` hook in the RCP integration, for the function that is throwing the 2 notices:
https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/integrations/class-rcp.php#L25

Judging by the `update_discount_affiliate` function, it's only supposed to be run from within the admin, but as we can see from the update function, the action hook also fires on the front-end of the website.

The 2 notices also appear when editing the RCP discount from within the admin, but only when a discount is saved without an affiliate added to the `Affiliate Discount?` field.

In addition to the notices, I found another bug whereby if you set an affiliate in the `Affiliate Discount?` field, but then try to change it to another affiliate and save, it retains the previous affiliate. Looking in the database I can see the `affwp_discount_rcp_` metakey being created twice for the same discount code. 

Will submit a PR in the next day or so.

pippinsplugins:
Good work tracking that down, it was confusing the heck out of me.

sumobi:
@pippinsplugins First pass: https://github.com/AffiliateWP/AffiliateWP/pull/1154

This PR fixes a few things:

1. The PHP notices on both the front-end when purchasing a membership using a discount code.

[...]

PR #1154, opened 2016-03-16
issue/1116 – rewrite update_discount_affiliate()

sumobi:
First pass at #1116

### DIFF
--- a/includes/integrations/class-rcp.php
+++ b/includes/integrations/class-rcp.php
@@ -219,22 +219,37 @@ class Affiliate_WP_RCP extends Affiliate_WP_Base {
 	*/
 	public function store_discount_affiliate( $args, $discount_id = 0 ) {
 
-		if( empty( $_POST['user_id'] ) && empty( $_POST['user_name'] ) ) {
+		if ( empty( $_POST['user_id'] ) && empty( $_POST['user_name'] ) ) {
 			return;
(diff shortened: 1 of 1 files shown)


includes/integrations/class-rcp.php
```diff
@@ -245,28 +245,61 @@ public function store_discount_affiliate( $args, $discount_id = 0 ) {
...
 		}
 
-		if( empty( $_POST['user_id'] ) && empty( $_POST['user_name'] ) ) {
-			return;
-		}
+		// Get the affiliate ID from the user ID
+		$affiliate_id = absint( affwp_get_affiliate_id( $user_id ) );
 
-		$affiliate_id = affwp_get_affiliate_id( $user_id );
+		// If an existing affiliate exists and a new affiliate is saved, delete the current meta key
```
pippinsplugins: No need to delete here since the update_user_meta will overwrite it anyway.

sumobi: It shouldn't because the update_user_meta is only updating the new `$user_id`, not the `$existing_user_id` that already exists in the database. This was the issue with the previous code, it added a new meta key since the `$user_id`'s are different.

pippinsplugins: Ah, missed that!

Issue #1097, opened 2016-02-11
Add helper function to dashboard.php


michaelbeil:
We should add a helper function to return the page ID with a filter on it for help with internationalization with plugins like WPML. 

That way we would not need to have users hardcode like this:

```php
	<?php
	/**
	 * Internationalise the affiliate-area page ID to output the correct link 
	 * with a page ID of 13492
	 * 
	 */
		$intl_affarea_pageID = apply_filters( 'wpml_object_id', 13492, 'page' );
	?>

	<li class="affwp-affiliate-dashboard-tab<?php echo $active_tab == 'urls' ? ' active' : ''; ?>">
		<a href="<?php echo get_permalink( $intl_affarea_pageID ); ?>?tab=urls"><?php _e( 'Affiliate URLs', 'affiliate-wp' ); ?></a>						</li>
	<li class="affwp-affiliate-dashboard-tab<?php echo $active_tab == 'stats' ? ' active' : ''; ?>">
		<a href="<?php echo get_permalink( $intl_affarea_pageID ); ?>?tab=stats"><?php _e( 'Statistics', 'affiliate-wp' ); ?></a>
	</li>
	<li class="affwp-affiliate-dashboard-tab<?php echo $active_tab == 'graphs' ? ' active' : ''; ?>">
		<a href="<?php echo get_permalink( $intl_affarea_pageID ); ?>?tab=graphs"><?php _e( 'Graphs', 'affiliate-wp' ); ?></a>
	</li>
	<li class="affwp-affiliate-dashboard-tab<?php echo $active_tab == 'referrals' ? ' active' : ''; ?>">
		<a href="<?php echo get_permalink( $intl_affarea_pageID ); ?>?tab=referrals"><?php _e( 'Referrals', 'affiliate-wp' ); ?></a>
	</li>
	<li class="affwp-affiliate-dashboard-tab<?php echo $active_tab == 'visits' ? ' active' : ''; ?>">
		<a href="<?php echo get_permalink( $intl_affarea_pageID ); ?>?tab=visits"><?php _e( 'Visits', 'affiliate-wp' ); ?></a>
	</li>
	<li class="affwp-affiliate-dashboard-tab<?php echo $active_tab == 'creatives' ? ' active' : ''; ?>">
		<a href="<?php echo get_permalink( $intl_affarea_pageID ); ?>?tab=creatives"><?php _e( 'Creatives', 'affiliate-wp' ); ?></a>
	</li>
[...]

PR #1161, opened 2016-03-29
Add filterable helpers for retrieving the Affiliates Area page ID and URL – #1097

DrewAPicture:
Significantly DRYs up templates/dashboard.php (yay!) in the process of allowing access to the page ID and URL.

Also adds some tests.

Issue: #1097


includes/affiliate-functions.php
```diff
@@ -1016,3 +1016,55 @@ function affwp_get_affiliate_base_url() {
 	return apply_filters( 'affwp_affiliate_referral_url_base', $base_url );
 
 }
+
+/**
+ * Retrieves the page ID for the Affiliates Area page.
+ *
+ * @since 1.8
+ *
+ * @return int Affiliates Area page ID.
```
pippinsplugins: We use singular `Affiliate` everywhere else. For consistency's sake, let's keep it singular.

includes/affiliate-functions.php
```diff
@@ -1016,3 +1016,55 @@ function affwp_get_affiliate_base_url() {
...
+	$affiliate_page_id = affiliate_wp()->settings->get( 'affiliates_page' );
+
+	/**
+	 * Filter the Affiliates Area page ID.
+	 *
+	 * @since 1.8
+	 *
+	 * @param int $affiliate_page_id Affiliates Area page ID.
+	 */
+	return apply_filters( 'affwp_affiliates_page_id', $affiliate_page_id );
```
pippinsplugins: Let's make this singular too.

Since it's theoretically possible that we might in the future introduce "affiliate account specific" pages, let's name the filter `affwp_affiliate_area_page_id`.

pippinsplugins: Actually, let's apply the same change to each of the helper function names too so that each one includes `affiliate_area_` in it.

Issue #1062, opened 2016-01-15
Cache flushing on referral insertion


mindctrl:
In the `add()` method of the `Affiliate_WP_Referrals_DB` class, the cache gets flushed after calling the `insert()` method in `Affiliate_WP_DB`. The latter already flushes the cache and there isn't a need to flush it again, is there?

Also the `insert()` method doesn't check for a successful insertion before flushing the cache. If it fails it'll still be flushed. Necessary?

https://github.com/AffiliateWP/AffiliateWP/blob/1.7.14/includes/class-referrals-db.php#L97

https://github.com/AffiliateWP/AffiliateWP/blob/1.7.14/includes/class-db.php#L85

pippinsplugins:
Whoops, that does seem wrong.

DrewAPicture:
PR: #1169

pippinsplugins:
Looks good!

PR #1169, opened 2016-03-31
Remove duplicative cache flush from the add() method in the referrals DB subclass

DrewAPicture:
Issue: #1062


DrewAPicture:
Note that 2b262c2 doesn't explicitly say it, but the second cache flush was removed in that commit.

Issue #999, opened 2015-12-04
Warning: number_format() expects parameter 1 to be double, string given


osthafen:
Hi,
I found a little warning-bug:

I's comming up with the $amount if the decimal or thousands separator are altered via str_replace:

Warning: number_format() expects parameter 1 to be double, string given in
 ../wp-content/plugins/affiliate-wp/includes/misc-functions.php on line 158

https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/misc-functions.php#L158

Greetings
Martin

pippinsplugins:
Hi Martin!

Could you show me an example of how you're altering it before the notice is displayed?

osthafen:
Hi Pippin,
I'm not altering it via code it's just in the settings page:
wp-admin/edit.php?post_type=download&page=edd-settings

sorry only german it's the german notation:
Tausender-Trennzeichen . Das Symbol, um die Tausender abzutrennen; gewöhnlich . (Punkt) oder , (Komma).
Zehner-Trennzeichen , Das Symbol, um die Zehner-Werte abzutrennen; gewöhnlich , (Komma) oder . (Punkt).

michaelbeil:
What should the German notation be @osthafen?

osthafen:
Everything is ok with the german notation. It's just the error coming up if you choose the german notation in the setting.
The function number_format() expects the parameter to be double but a string is provided. It's displaying this error warning int the overview page: /admin.php?page=affiliate-wp
'Warning: number_format() expects parameter 1 to be double, string given in ......./httpdocs/wp-content/plugins/affiliate-wp/includes/misc-functions.php on line 158'
![image](https://cloud.githubusercontent.com/assets/525826/11603877/1193bcae-9ae6-11e5-84bf-8ce2834d9d86.png)

pippinsplugins:
Thanks, we'll get that fixed.

DrewAPicture:
Added a PR: #1170.

Includes a test for the float value and fixes an unrelated undefined notice triggered when testing commas as decimal separators.

PR #1170, opened 2016-03-31
Keep a float afloat when formatting amounts with thousands separators

DrewAPicture:
Issue: #999

Includes a test.

### DIFF
--- a/includes/admin/settings/class-settings.php
+++ b/includes/admin/settings/class-settings.php
@@ -67,6 +67,18 @@ class Affiliate_WP_Settings {
 	}
 
 	/**
+	 * Sets an option (in memory).
+	 *
+	 * @since 1.8
+	 *
+	 * @param string $key   The option key to set.
+	 * @param mixed  $value The value to assign to the key.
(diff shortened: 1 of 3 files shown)


includes/admin/settings/class-settings.php
```diff
@@ -67,6 +67,19 @@ public function get( $key, $default = false ) {
...
 	/**
+	 * Sets an option (in memory).
+	 *
+	 * @since 1.8
+	 * @access public
+	 *
+	 * @param string $key   The option key to set.
+	 * @param mixed  $value The value to assign to the key.
+	 */
+	public function set( $key, $value ) {
```
pippinsplugins: Since this only sets it in memory, we should probably also consider a `save()` method to store temp changes in the database.

DrewAPicture: OK.

DrewAPicture: For reference, opened #1171 for discussion on a `save()` method.

Issue #1171, opened 2016-03-31
Standardize setting and saving settings on the fly


DrewAPicture:
In #999 I've added a `Affiliate_WP_Settings::set()` method to better facilitate testing, and @pippinsplugins suggested pairing that with a `save()` method on the related PR. After some discussion, we decided this should be its own task.

Much like `EDD_Payments`, which uses a `$pending` property to effectively queue settings changes, we can do the same thing for AWP settings, which will help us reduce code duplication in various places, while simultaneously simplifying and standardizing best-practice for saving AWP settings.

In [install.php](../blob/master/includes/install.php#L38), for instance, here's what that might look like:

**Current:**
```php
$options = $affiliate_wp_install->settings->get_all();
$options['affiliates_page'] = $affiliate_area;
update_option( 'affwp_settings', $options );
```

**Queuing via `set()` and saving via `save()`:**
```php
$affiliate_wp_install->settings->set( 'affiliates_page', $affiliate_area );
$affiliate_wp_install->settings->save();
```

**Queuing via `set()` and saving in the same step with an optional third parameter:**
```php
$affiliate_wp_install->settings->set( 'affiliates_page', $affiliate_area, $save = true );
```

**A third option would be to allow chaining with the `set()` method, e.g.:**
```php
$affiliate_wp_install->settings->set( 'affiliates_page', $affiliate_area )->save();
```

Either way, the addition of the `set()` method in #999 only works in memory. A `save()` method obviously has the ability to literally save that value, so we'd probably want to implement a whitelist based on register setting keys or similar to prevent misuse.

pippinsplugins:
I prefer this option:
```php
$affiliate_wp_install->settings->set( 'affiliates_page', $affiliate_area, $save = true );
```

DrewAPicture:
Added a PR: #1172 

Tests incoming.

pippinsplugins:
This is really excellent. Merging as is.

pippinsplugins:
I've tested this quite a bit. @sumobi @michaelbeil could you guys run through some quick tests too just to make sure no edge case was missed? It's merged into `release/1.8`.

michaelbeil:
This is great. Did a once over and the settings area still works :smile:. Saving settings is super speedy!

sumobi:
All looks good to me

PR #1172, opened 2016-04-01
Implement saving settings via the Affiliate_WP_Settings::set() method

DrewAPicture:
Issue: #1171
Related discussion: #999 

The new `Affiliate_WP_Settings::set()` method accepts an array of setting `key => value` pairs and an optional `save()` trigger.

I've implemented the `save()` method as protected, currently only to be called via the optional third parameter in `set()`. We can always switch this to public later if the need arises.

**Basic single setting usage:**
```php
		$affiliate_area = wp_insert_post(
			array(
				'post_title'     => __( 'Affiliate Area', 'affiliate-wp' ),
				'post_content'   => '[affiliate_area]',
				'post_status'    => 'publish',
				'post_author'    => 1,
				'post_type'      => 'page',
				'comment_status' => 'closed'
			)
		);

		// Update settings.
		affiliate_wp()->settings->set( array(
			'affiliates_page' => $affiliate_area
		), $save = true );
```

**Basic array of settings usage:**
```php
	// Retrieve the settings from the file and convert the json object to an array
	$settings = affwp_object_to_array( json_decode( file_get_contents( $import_file ) ) );

	// Update settings.
	affiliate_wp()->settings->set( $settings, $save = true );
```

### DIFF
--- a/includes/admin/class-upgrades.php
+++ b/includes/admin/class-upgrades.php
@@ -194,8 +194,8 @@ class Affiliate_WP_Upgrades {
 			$settings['referral_rate'] = floatval( $rate );
 		}
 
-		update_option( 'affwp_settings', $settings );
-
+		// Update settings.
+		affiliate_wp()->settings->set( $settings, null, $save = true );
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


DrewAPicture:
Tests are incoming.

PR #1183, opened 2016-04-05
add default_url to shortcode affiliate_referral

areski:


michaelbeil:
@areski just curious, how does this help you?

areski:
@michaelbeil actually I think that is not helping me.
What I would like to do, it's customer come on our website www.company.com?ref=1
and we can send them out on our other website with the same ref parameter and create a link to www.otherwebsite.com?ref=1

`1shortcode affiliate_referral` only work for logged affiliate I would like something that work for visitors, any ideas how to achieve this? should I create a new shortcode?

michaelbeil:
So, you are wanting to create an affiliate link with a different URL where AffiliateWP is not installed?

areski:
exactly, we have an other store where we process the payment and I want to track the affiliate there.
So I would to be able to create a link and add the affiliate `ref` ID on it

areski:
the last commit seems to solve it for me!

michaelbeil:
I do not know that we would want to include this in the core plugin. You could make your own extension to accomplish what you are wanting, if you like.

areski:
yes I agree, an other plugin could do which simple build shortcode on a cookie value.
I was curious that nobody had the same usecase as me and need to send the referral to an other website.

I'm closing the PR

michaelbeil:
We also have an external referral links add-on that allows you to pass the ref parameter: https://affiliatewp.com/addons/external-referral-links.

areski:
Thanks, I wish I would have find it earlier :)

PR #1193, opened 2016-04-13
Make affiliate status string translatable. Fixes #997.

### LINKED ISSUE #997: Fix non-translatable affiliate status string
In:
```
<?php if( $affiliates ) : ?>
    <?php foreach( $affiliates as $affiliate  ) : ?>    
        <tr>
            <td><?php echo affiliate_wp()->affiliates->get_affiliate_name( $affiliate->affiliate_id ); ?></td>
            <td><?php echo $affiliate->status; ?></td>
            <td>
                <?php
                if( 'pending' == $affiliate->status ) {
```
https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/admin/overview/overview.php#L124-L131 

We should allow $affiliate->status to be translatable.
 
On the overview page, the "active/pending" translation is not being displayed, nor is it for any translation since we query the DB. 

```
# @ affiliate-wp
#: includes/admin/affiliates/affiliates.php:204
#: includes/admin/affiliates/new.php:42
#: includes/admin/creatives/creatives.php:143
#: includes/admin/creatives/edit.php:99 includes/admin/creatives/new.php:89
#: includes/admin/tools/tools.php:230
msgid "Active"
msgstr "Ativo"
```
  
Grunt, of course, will not pick it up: http://beil.io/QKXa. 
 
Is there a translation function that we might be able to use, and translate each possible phrase? 
 
@sumobi said:
> we could just go IF $ status = active, then display the active test string, and the same for the other within `<?php echo $affiliate->status; ?>`.

ramiabraham:
Fixes #997


ramiabraham:
(closing and re-submitting against release/1.8)

PR #1196, opened 2016-04-13
Make affiliate status string translatable, related test (fixes #997)

### LINKED ISSUE #997: Fix non-translatable affiliate status string

ramiabraham:
Make affiliate status string translatable, related test (fixes #997)

### DIFF
--- a/includes/affiliate-functions.php
+++ b/includes/affiliate-functions.php
@@ -222,9 +222,7 @@ function affwp_get_affiliate_status_label( $affiliate ) {
 			break;
 	}
 
-	$status_label = apply_filters( 'affwp_get_affiliate_status_label' );
-
-	return $status_label;
+	return apply_filters( 'affwp_get_affiliate_status_label', $status_label );
 }
 
 /**
--- a/includes/affiliate-functions.php
+++ b/includes/affiliate-functions.php
@@ -185,6 +185,44 @@ function affwp_set_affiliate_status( $affiliate, $status = '' ) {
 }
 
 /**
+ * Retrieves the affiliate's status and returns a translatable string
+ *
+ * @since 1.8
+ * @param  $status_label A translatable, filterable label indicating affiliate status
+ * @return string
+ */
+function affwp_get_affiliate_status_label( $affiliate ) {
+
+	$affiliate    = affwp_get_affiliate( $affiliate );
+	$status       = '';
+	$status_label = '';
+
+	// Get current affiliate status
+	$status       = $affiliate->status;
+
+	// Return translatable string
+	switch( $status ) {
+
+		case 'active':
+			$status_label = __( 'Active','affiliate-wp' );
+			break;
+		case 'inactive':
+			$status_label = __( 'Inactive','affiliate-wp' );
+			break;
+		case 'pending':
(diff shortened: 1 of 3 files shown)


includes/affiliate-functions.php
```diff
@@ -185,6 +185,44 @@ function affwp_set_affiliate_status( $affiliate, $status = '' ) {
...
 /**
+ * Retrieves the affiliate's status and returns a translatable string
+ *
+ * @since 1.8
+ * @param  $status_label A translatable, filterable label indicating affiliate status
+ * @return string
+ */
+function affwp_get_affiliate_status_label( $affiliate ) {
+
+	$affiliate    = affwp_get_affiliate( $affiliate );
```
pippinsplugins: Let's have a check here to make sure we actually returned an affiliate successfully.

includes/affiliate-functions.php
```diff
@@ -185,6 +185,44 @@ function affwp_set_affiliate_status( $affiliate, $status = '' ) {
 }
 
 /**
+ * Retrieves the affiliate's status and returns a translatable string
+ *
+ * @since 1.8
+ * @param  $status_label A translatable, filterable label indicating affiliate status
```
pippinsplugins: @param is incorrect here. It should be `$affiliate`. The `$status_label` is the return.

includes/affiliate-functions.php
```diff
@@ -185,6 +185,44 @@ function affwp_set_affiliate_status( $affiliate, $status = '' ) {
...
+			break;
+		case 'pending':
+			$status_label = __( 'Pending','affiliate-wp' );
+			break;
+		case 'rejected':
+			$status_label = __( 'Rejected','affiliate-wp' );
+			break;
+	}
+
+	apply_filters( 'affwp_get_affiliate_status_label', $status_label, $status );
```
pippinsplugins: Should be `$status_label = apply_filters` or just `return apply_filters`

PR #1208, opened 2016-04-18
Define Affiliate_WP::below_php_version_notice() as a static method. Fixes #1201 

### LINKED ISSUE #1201: Affiliate_WP::below_php_version_notice() should be a static method
`below_php_version_notice` method in `affiliate-wp.php:246` is called statically but not declared as such and throws a notice by strict standards.

How to replicate:
* change compared version (5.3) in `affiliate-wp.php:195` to something higher than your PHP version to trigger the `admin_notices` action

ramiabraham:
Define Affiliate_WP::below_php_version_notice() as a static method. Fixes #1201


pippinsplugins:
@ramiabraham you'll need to base your branch off of an up-to-date `release/1.8` and then push that up. That will give you just the single commit in the PR.

Issue #1182, opened 2016-04-04
Add default statement to notices class


sumobi:
Related: https://github.com/AffiliateWP/AffiliateWP/issues/1078

Any reason why we can't add a `default` statement here?
https://github.com/AffiliateWP/AffiliateWP/blob/release/1.8/includes/admin/class-notices.php#L226

I've just added an admin notice to the `Allowed Products` add-on but it won't be able to save the user meta once 1.8 is released, since it doesn't match any of the notices allowed.

pippinsplugins:
@sumobi Nope. Let's add a `default` with a `do_action`.

DrewAPicture:
Added a PR: #1212

pippinsplugins:
Merged.

PR #1212, opened 2016-04-19
Issue/1182 – Add a default statement to notices class + new action

DrewAPicture:
Issue: #1182 

Adds a default case to the switch statement in `Affiliate_WP_Admin_Notices::dismiss_notices()` along with a new action.

### DIFF
--- a/includes/admin/class-notices.php
+++ b/includes/admin/class-notices.php
@@ -231,6 +231,16 @@ class Affiliate_WP_Admin_Notices {
 				case 'invalid_license':
 					set_transient( 'affwp_license_notice', true, 2 * WEEK_IN_SECONDS );
 					break;
+				default:
+					/**
+					 * Fires once a notice has been flagged for dismissal.
+					 *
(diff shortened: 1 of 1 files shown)


includes/admin/class-notices.php
```diff
@@ -230,6 +231,16 @@ function dismiss_notices() {
...
 					break;
+				default:
+					/**
+					 * Fires once a notice has been flagged for dismissal.
+					 *
+					 * @since 1.8
+					 *
+					 * @param string $notice Notice value via $_GET['affwp_notice'].
+					 */
+					do_action( 'dismiss_notices', $notice );
```
sumobi: @pippinsplugins @DrewAPicture Let's prefix this with `affwp_`

DrewAPicture: Yeah, good call :)

DrewAPicture:
@sumobi The action has been prefixed in 12db3c20da and pushed to release/1.8 directly.

Issue #1224, opened 2016-04-22
Unable to sort "Unpaid Referrals" column correctly


sumobi:
To replicate:

1. Go to Affiliates -> Affiliates.

2. Click on the `Unpaid Referrals` column title to try and sort the unpaid referrals.

3. Expected result: to have the referrals show in order. What it's actually doing is sorting the `Affiliate ID` column, instead of the unpaid referrals.

DrewAPicture:
Added a PR: #1230

DrewAPicture:
Note: #1230 also fixes similar incorrect sorting behavior for the Name column. See the PR for more on that.

pippinsplugins:
@DrewAPicture Everything looks good here, though we could avoid the count query when ordering by visits since we keep a count in the `visits` column of the affiliates table already.

DrewAPicture:
I've removed the extra visits logic. Good to know.

michaelbeil:
Confirmed fix on branch `issue/1224`.

PR #1230, opened 2016-04-24
issue/1224 – Fix sorting for the Name, Username, Unpaid and Paid Referrals columns for Affiliates

DrewAPicture:
Issue: #1224 

This PR fixes sorting on Affiliates > Affiliates for the following columns:

* Name
* Username
* Unpaid Referrals
* Paid Referrals

Background:
As originally reported in #1224, the sortable **Unpaid Referrals** column on Affiliates > Affiliates previously appeared to be ordering the table by `affiliate_id` rather than unpaid referrals count.

After some investigation, it turns out that logic was actually never added to handle referral statuses as orderby values, thus the orderby column was falling back to `affiliate_id`. This PR fixes ordering by both the **Unpaid Referrals** and **Paid Referrals** columns,

Similarly, when investigating Unpaid Referrals, I also noticed that the sortable **Name** (user display name) and **Username** columns were exhibiting the same behavior: sorting the table by `affiliate_id` because of incorrect or missing orderby logic.

This PR includes tests for all possible orderby permutations (minus username and visits).


includes/class-affiliates-db.php
```diff
@@ -217,19 +220,53 @@ public function get_affiliates( $args = array(), $count = false ) {
...
+
+			case 'earnings':
+				// Earnings.
+				$orderby = 'earnings+0';
+				break;
+
+			case 'visits':
+				// If ordering by visits, do a sub-query based on the count.
+				$visits = affiliate_wp()->visits->table_name;
+
```
pippinsplugins: The visit count of an affiliate is stored in the `visits` column of the affiliates database, so no need to run the extra query here.

DrewAPicture:
@pippinsplugins: OK, removed the visits case. I wasn't sure how synced the vote count in the affiliates table was vs the actual visits table.

pippinsplugins:
It's synced each time a visit is recorded.

Issue #1243, opened 2016-04-27
Support removal of Affiliate Area tabs


sumobi:
https://github.com/AffiliateWP/affiliatewp-affiliate-area-tabs/issues/1

sumobi:
First pass. Looks like we might be able to clean up the template file a bit more.

pippinsplugins:
I'm not a huge fan of calling the same filter over and over. How about a helper function instead? Something like `affwp_should_show_tab( $tab )`? That can include a filter.

sumobi:
Good idea

sumobi:
@pippinsplugins check latest commit. Simplified the overall `dashboard.php` template and introduced some new helper functions. Winning.

pippinsplugins:
@sumobi Can we move this to 1.8? Tentatively looking at a 1.8 release next week.

sumobi:
Yep, I'll resubmit against 1.8

sumobi:
@pippinsplugins Nice, I was not aware of the new functions in 1.8 for getting the page ID and URL of the affiliate area.

Any reason why we've added this variable? Going to remove it otherwise
https://github.com/AffiliateWP/AffiliateWP/blob/release/1.8/templates/dashboard.php#L4

pippinsplugins:
@sumobi none that I see. Kill it.

pippinsplugins:
@sumobi I'd like to see `affwp_affiliate_area_tab( $tab )` renamed to `affwp_affiliate_area_show_tab( $tab )` to make it easier to grok what the function does.

DrewAPicture:
FWIW, I think the `$affiliates_page_id = affwp_get_affiliate_area_page_id();` line was leftover from before we introduced `affwp_get_affiliate_area_page_url()`. So yeah, kill it :+1:

sumobi:
Ok think we're good to go on this

pippinsplugins:
Looks good to me.

PR #1244, opened 2016-04-27
issues/1243 - Introduce affwp_dashboard_tabs filter

sumobi:
#1243

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -515,7 +515,7 @@ function affwp_tab( $tab = '', $label = '' ) {
 	$active_tab = ! empty( $_GET['tab'] ) ? sanitize_text_field( $_GET['tab'] ) : 'urls';
 
 	if ( affwp_tabs_show_tab( $tab ) ) : ?>
-	<li class="affwp-affiliate-dashboard-tab<?php echo affwp_tabs_active_class( $active_tab, $tab ); ?>">
+	<li class="affwp-affiliate-dashboard-tab <?php echo $tab; ?><?php echo affwp_tabs_active_class( $active_tab, $tab ); ?>">
 		<a href="<?php echo affwp_tabs_get_url( $tab ); ?>"><?php echo $label; ?></a>
 	</li>
 	<?php endif;
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -538,7 +538,7 @@ function affwp_tabs_show_tab( $tab = '' ) {
  * @return string active CSS class
  */
 function affwp_tabs_active_class( $active_tab, $tab = '' ) {
-	$class = apply_filters( 'affwp_tabs_active_class', 'active' );
+	$class = apply_filters( 'affwp_tabs_active_class', 'active', $tab );
 	return $active_tab == $tab ? ' ' . $class : '';
 }
 
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -493,3 +493,61 @@ function affwp_abs_number_round( $val, $precision = 2 ) {
 	return $val;
 
 }
+
+/**
+ * Retrieves the Affiliate Area URL
+ *
+ * @since  1.7.18
+ * @return string URL to the Affiliate Area
+ */
+function affwp_get_affiliate_area_url() {
+	return apply_filters( 'affwp_affiliate_area_url', get_permalink( affiliate_wp()->settings->get( 'affiliates_page' ) ) );
+}
+
+/**
+ * Add a new tab
+ *
+ * @since  1.7.18
+ * @return string
+ */
+function affwp_tab( $tab = '', $label = '' ) {
+
+	$active_tab = ! empty( $_GET['tab'] ) ? sanitize_text_field( $_GET['tab'] ) : 'urls';
+
+	if ( affwp_tabs_show_tab( $tab ) ) : ?>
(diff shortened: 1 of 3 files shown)


includes/misc-functions.php
```diff
@@ -493,3 +493,61 @@ function affwp_abs_number_round( $val, $precision = 2 ) {
...
+	return apply_filters( 'affwp_affiliate_area_url', get_permalink( affiliate_wp()->settings->get( 'affiliates_page' ) ) );
+}
+
+/**
+ * Add a new tab
+ *
+ * @since  1.7.18
+ * @return string
+ */
+function affwp_tab( $tab = '', $label = '' ) {
```
pippinsplugins: Let's name this so it's specific to frontend / affliate area.

includes/misc-functions.php
```diff
@@ -493,3 +493,61 @@ function affwp_abs_number_round( $val, $precision = 2 ) {
...
 }
+
+/**
+ * Retrieves the Affiliate Area URL
+ *
+ * @since  1.7.18
+ * @return string URL to the Affiliate Area
+ */
+function affwp_get_affiliate_area_url() {
+	return apply_filters( 'affwp_affiliate_area_url', get_permalink( affiliate_wp()->settings->get( 'affiliates_page' ) ) );
```
pippinsplugins: I think we added a function for retrieving affiliate area page ID. Can we use that please?

includes/misc-functions.php
```diff
@@ -493,3 +493,61 @@ function affwp_abs_number_round( $val, $precision = 2 ) {
...
+}
+
+/**
+ * Outputs the active CSS class for each tab in the Affiliate Area
+ *
+ * @since  1.7.18
+ * @return string active CSS class
+ */
+function affwp_tabs_active_class( $active_tab, $tab = '' ) {
+	$class = apply_filters( 'affwp_tabs_active_class', 'active' );
```
pippinsplugins: The tab needs to be passed to the filter.

includes/misc-functions.php
```diff
@@ -493,3 +493,61 @@ function affwp_abs_number_round( $val, $precision = 2 ) {
...
+ */
+function affwp_tab( $tab = '', $label = '' ) {
+
+	$active_tab = ! empty( $_GET['tab'] ) ? sanitize_text_field( $_GET['tab'] ) : 'urls';
+
+	if ( affwp_tabs_show_tab( $tab ) ) : ?>
+	<li class="affwp-affiliate-dashboard-tab<?php echo affwp_tabs_active_class( $active_tab, $tab ); ?>">
+		<a href="<?php echo affwp_tabs_get_url( $tab ); ?>"><?php echo $label; ?></a>
+	</li>
+	<?php endif;
```
pippinsplugins: I'm not sure I love the tab HTML being in a function. That makes it significantly harder for someone to customize the HTML, even for really simply tasks like adding icons.

templates/dashboard.php
```diff
@@ -29,35 +29,33 @@
...
+
+			<?php echo affwp_tab( 'creatives', __( 'Creatives', 'affiliate-wp' ) ); ?>
+
+			<?php echo affwp_tab( 'settings', __( 'Settings', 'affiliate-wp' ) ); ?>
+
 			<?php do_action( 'affwp_affiliate_dashboard_tabs', affwp_get_affiliate_id(), $active_tab ); ?>
 		</ul>
 
-		<?php affiliate_wp()->templates->get_template_part( 'dashboard-tab', $active_tab ); ?>
+		<?php
```
pippinsplugins: This needs to use the helper function.

includes/misc-functions.php
sumobi: We do, but it's a function specific to the login class. I've changed `get_login_url` to use the new function here: https://github.com/AffiliateWP/AffiliateWP/blob/issue/1243/includes/class-login.php#L167

This way you can filter the individual login URL separately. There are times for example where you might want to login to a different URL.

includes/misc-functions.php
sumobi: Ok

includes/misc-functions.php
sumobi: Good catch

includes/misc-functions.php
sumobi: Hmm good point... Should we kill off the new affwp_tab() function and move this back to the template file?

sumobi: Or add a filter to this function so you can filter the HTML based on the tab? This way you can still add icons etc, and won't need to edit the entire dashboard.php function to do so. We can provide an example of how to do so.

includes/misc-functions.php
sumobi: done

includes/misc-functions.php
sumobi: done

templates/dashboard.php
sumobi: done

includes/misc-functions.php
pippinsplugins: For ease of customization, I think it's important as much of the HTML as possible be in the template files, even if it means a bit of duplication. That is the point of template files after all.

sumobi: 👍  removed

michaelbeil:
This was added in https://github.com/AffiliateWP/AffiliateWP/pull/1249.

Issue #1119, opened 2016-03-01
LifterLMS Integration


michaelbeil:
Proposal: add LifterLMS integration.

Some info: 208 `do_action`s and 268 `apply_filters` are available.

pippinsplugins:
Milestoned for 1.5.8.

pippinsplugins:
1.8.3 actually.

pippinsplugins:
@ramiabraham could you please test this integration tomorrow? If all is working, please also go ahead and write up the documentation page for it.

ramiabraham:
Looks good here.

- LifterLMS coupons are attributed to the correct AffiliateWP affiliate. 
- Aside from the noted js-related adjustment made in https://github.com/AffiliateWP/AffiliateWP/commit/89a0895d2fccee6c95750c9c4bfb737ab9ebec9a, there were no other required adjustments.

Will update when doc is completed for review.

michaelbeil:
Looking good. Just made a few comments on the latest commit.

pippinsplugins:
Just pushed a couple of minor changes to https://github.com/AffiliateWP/AffiliateWP/pull/1368. Once tests pass, we can merge.

michaelbeil:
Looks good.

michaelbeil:
New PR: #1368

PR #1257, opened 2016-05-03
lifterlms integration

thomasplevy:
Proposed LifterLMS Integration as per issue #1119 

A few things to note:

The current stable version of LifterLMS is 2.6.3 and in that release we added an action hook that we're hooking into to kick the whole thing off. This is `lifterlms_order_complete`, we have a few older actions that we *could* use but nothing that was passing order related information so we could reliably get the reference. We decided to just add an action to keep the amount of querying low inside AffiliateWP.

That said, should we worry about doing a version compare inside AffiliateWP when the integration is enabled? Is there any warning methods available in AffiliateWP to note that version upgrade requirement? We'll, of course, notify our users who are aching to get this built that they need this version to use the integration but I'm concerned about anyone who organically stumbles across the integration and clicks the button.

We've also included some inline JS that would run on LifterLMS Courses & Memberships to make the UI a little more pleasant for our users. This isn't necessary if you're opposed to the inline JS included in this PR. I just wanted to note that in case it looks extremely offensive to you please let me know and we'll rip it out and resubmit. I'm not a huge fan of the inline JS but I am a fan of the UX it provides. I didn't see precedence for adding a whole JS file for the integration (and frankly that doesn't feel necessary if the inline is acceptable to ya'll).

Thanks so much!

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -240,7 +240,7 @@ jQuery(document).ready(function($) {
 	window.llms.metabox_product_affwp = function() {
 
 		/**
-		 * Initialize and Bind events
+		 * Initialize and Bind events if our check element is found
 		 * @return void
 		 */
 		this.init = function() {
@@ -248,13 +248,16 @@ jQuery(document).ready(function($) {
 			// only bind if our hidden input exists in the dom
 			if ( $( '#affwp-llms-enabled' ).length ) {
 
-				console.log( 'init' );
 				this.bind();
 
 			}
 
 		};
 
+		/**
+		 * Bind dom events
+		 * @return void
+		 */
 		this.bind = function() {
 
 			this.bind_disable_field();
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -228,3 +228,105 @@ jQuery(document).ready(function($) {
 	});
 
 });
+
+( function( $ ) {
+
+	window.llms = window.llms || {};
+
+	/**
+	 * Handle the AffiliateWP Tab JS interaction
+	 * @return obj
+	 */
+	window.llms.metabox_product_affwp = function() {
+
+		/**
+		 * Initialize and Bind events
+		 * @return void
+		 */
+		this.init = function() {
+
+			// only bind if our hidden input exists in the dom
+			if ( $( '#affwp-llms-enabled' ).length ) {
+
+				console.log( 'init' );
+				this.bind();
+
+			}
+
+		};
+
+		this.bind = function() {
+
+			this.bind_disable_field();
+			$( '#_affwp_disable_referrals' ).trigger( 'change' );
+
+			this.bind_override_field();
+			$( '#_affwp_enable_referral_overrides' ).trigger( 'change' );
+
+		};
+
+		/**
+		 * Bind thie "disable referrals" fields
+		 * @return void
+		 */
+		this.bind_disable_field = function() {
+
+			$( '#_affwp_disable_referrals' ).on( 'change', function() {
+
+				var $group = $( '.llms-affwp-disable-fields');
+
+				if ( $(this).is( ':checked' ) ) {
+
+					$group.hide( 200 );
+					$( '#_affwp_enable_referral_overrides' ).removeAttr( 'checked' ).trigger( 'change' );
+
+				} else {
+
+					$group.show( 200 );
+
+				}
+
+			} );
+
+		};
+
+		/**
+		 * Bind the "enable overrides" field
+		 * @return void
+		 */
+		this.bind_override_field = function() {
+
+			$( '#_affwp_enable_referral_overrides' ).on( 'change', function() {
+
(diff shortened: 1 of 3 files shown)


includes/integrations/class-lifterlms.php
```diff
@@ -0,0 +1,725 @@
...
+
+		// if an existing referral exists and it is paid or unpaid exit.
+		if ( $existing && ( 'paid' === $existing->status || 'unpaid' === $existing->status ) ) {
+
+			return;
+
+		}
+
+		// if this was a referral or we have a coupon and a coupon affiliate id
+		if ( $this->was_referred() || ( $order->coupon && $order->coupon_affiliate_id ) ) {
```
pippinsplugins: Let's put this check at the very top. No need to run any logic if the customer wasn't referred.

includes/integrations/class-lifterlms.php
```diff
@@ -0,0 +1,725 @@
...
+
+			// get the referring affiliate's affiliate id
+			$affiliate_id = $this->get_affiliate_id( $order_id );
+
+			// use our coupon affiliate if we have one
+			// @note 	is this the correct way to do this? Always override with the coupon?
```
pippinsplugins: Yes, coupon always takes precedence.

includes/integrations/class-lifterlms.php
```diff
@@ -0,0 +1,725 @@
...
+				return;
+			}
+
+
+			$description = apply_filters( 'affwp_llms_get_referral_description', $order->product_title, $order, $affiliate_id );
+			$visit_id = affiliate_wp()->tracking->get_visit_id();
+
+			// update existing referral if it exists
+			// this isn't currently ever going to happen with LifterLMS but leaving it here for future use
+			if ( $existing ) {
```
pippinsplugins: Since $existing isn't defined, it will need to be defined or this logic removed.

pippinsplugins:
Looks really good!

I don't mind the inline JS but let's go ahead and move it to our admin JS file so that it's combined with the rest of our JS. Any issues doing that?

includes/integrations/class-lifterlms.php
thomasplevy: 👍

includes/integrations/class-lifterlms.php
thomasplevy: thanks

includes/integrations/class-lifterlms.php
thomasplevy: $existing is defined on line 91 (https://github.com/AffiliateWP/AffiliateWP/pull/1257/files#diff-8893037b472760bac3785d4bbd55b49dR91)

thomasplevy:
@pippinsplugins no problem re the JS I just didn't want to be so arrogant as to add some JS to one of your files.

Will add to the JS and run a few more tests and update the fork in a little while

pippinsplugins:
Great, thanks!

thomasplevy:
@pippinsplugins I can't see any obvious minification task in the gruntfile or related directories, how do you handle minification?

thomasplevy:
moved inline js out of class-lifterlms.php and fiddled with the JS to ensure nothing gets funky. As noted above I don't see any JS minification tasks in the gruntfile so the JS is not minified, enabled SCRIPT_DEBUG to do some testing on the unminified. Also added a check so that llms related JS will only bind on pages it's needed see the init function.

Please let me know if you see any other issues with the integration and thanks again!

pippinsplugins:
We minify before each release so nothing to worry about there!

pippinsplugins:
We'll get this reviewed and tested further shortly after our 1.8 release, which is scheduled for Thursday this week.

assets/js/admin.js
```diff
@@ -228,3 +228,108 @@ jQuery(document).ready(function($) {
...
+		// go
+		this.init();
+
+		// return, just bc
+		return this;
+
+	};
+
+	// instatiate the class
+	var a = new window.llms.metabox_product_affwp();
```
ramiabraham: Th AffiliateWP `admin.js` is for items relating to AffiliateWP core, or _all_ integrations, and cannot contain code relating to a single integration.

Will move this to the `class-lifterlms.php` file as inline js for now, however, allowing integrations to define their own assets is something we're considering for a future update.

ramiabraham:
@thomasplevy Thanks for your work here! We're going to get this into AffiliateWP `1.8.3`. One adjustment is that we chose to move the js back inline. It's defined in the method `inline_js` within the LMMS integration class.

thomasplevy:
@ramiabraham,

I actually think this is better given the current architecture, I hate the idea of bloating (however small) your entire JS file for users who have no interest in LifterLMS.

Thanks for the update and please let me know if you need anything from me!

pippinsplugins:
Implemented through https://github.com/AffiliateWP/AffiliateWP/pull/1368

Issue #1184, opened 2016-04-05
Core objects for affiliates, referrals, creatives, and visits


DrewAPicture:
In preparation for a possible REST endpoints in 1.9, we should switch to using a legitimate objects for affiliates, referrals, creatives, and visits over `stdClass`. This would vastly simplify any testing we have in place as well as code leveraging IDs or objects everywhere.

The other upside of having predictably-structured objects is exposure via REST endpoints or WP-CLI command support down the line would be instantly reflected without having to do things twice.

DrewAPicture:
Related: #1267

DrewAPicture:
For some perspective on what this looks like in memcached, here's the hit/miss table. Hint: it's working:
![Memcached hit/miss](http://cl.ly/1O1F0Y100u2Q/Screen%20Shot%202016-05-21%20at%209.31.16%20AM.png)

In monitoring the live stats when loading queries such the Affiliates > Affiliates screen et al or performing CRUD operations, we're consistently seeing SETs spiking as expected at the time of change then cratering out on the second load along with high HIT percentages. All good signs :)

DrewAPicture:
I've split implementation into a new issue/PR pair, #1301 and #1302 respectively.

pippinsplugins:
Looking really good so far. Love seeing the `set()` and `save()` methods.

I did just notice that the `custom` is doc'ed with the wrong data type:
```php
/**
 * Custom referral data.
 *
 * @since 1.9
 * @access public
 * @var string
 */
public $custom;
```

It's mixed. It can be a string, int, or array depending on what the integration chooses to do with it.

pippinsplugins:
```php
/**
 * Products associated with the referral.
 *
 * @since 1.9
 * @access public
 * @var string
 */
public $products;
```
It's an array, not a string.

pippinsplugins:
I'm also seeing a strict standards notice:

```
Strict standards: Static function AffWP_Object::sanitize_field() should not be abstract in /Users/pippinwilliamson/Sites/trunk/wp-content/plugins/affiliate-wp/includes/objects/class-affwp-object.php on line 230
```

pippinsplugins:
Something else weird I'm seeing is that retrieving objects isn't working. All values remain default instead of being filled from the values in the DB.

DrewAPicture:
@pippinsplugins Can you supply some sample object retrieval code so I can try to reproduce?

pippinsplugins:
[...]

PR #1265, opened 2016-05-05
issue/1184 – Establish core objects for affiliates, referrals, creatives, and visits

DrewAPicture:
Issue: #1184 

This PR only handles introducing the new object classes. For implementation, see #1301 (PR: #1302).


includes/objects/class-affwp-affiliate.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+final class AffWP_Affiliate extends AffWP_Object {
```
pippinsplugins: Let's plan to keep `final` for a version or two and then remove it once we're sure it's safe to extend.

includes/objects/class-affwp-affiliate.php
```diff
@@ -0,0 +1,135 @@
...
+	 * Object type.
+	 *
+	 * Used as the cache group and for accessing object DB classes in the parent.
+	 *
+	 * @since 1.9
+	 * @access public
+	 * @static
+	 * @var string
+	 */
+	public static $object_type = 'affiliates';
```
pippinsplugins: Can we make this singular?

pippinsplugins: Same with the other object ypes.

includes/objects/class-affwp-affiliate.php
DrewAPicture: Roger.

Issue #1229, opened 2016-04-24
License Key: Make configurable via wp-config.php


JanThiel:
Hi,

when running AffiliateWP in a Multisite, you need to set the License Key on each site manually. If you want the updates to work, it also has to be activated on the Networks main page. Although if you don't wanna use it there. Further more when using automatic deployments, each UI input for license keys is breaking scalability.

Simple Solution: Offer an alternative way to set the license key in wp-config.php instead of UI + this->options.
So check for is_defined("wpaff_license") before checking the this->settings->license_key and continue stuff with that value instead.

In the Admin Backend the key should not be visible or editable. Just to make clear it is set at another place and not there. As it cannot and should not be edited in the WP-Admin backend.

Thanks, Best Regards,

Jan

pippinsplugins:
Yes! Let's do it.

DrewAPicture:
Added a PR: #1268

JanThiel:
@DrewAPicture There are two more things I would like to throw into for consideration. I also did some implementation work on this one (your PR made my code obsolete, which is fine anyway) so I would like to share my thoughts:

1. The License Key Section in the wp-admin backend should be disabled when the code is set via define( ... ). I implemented a "disabled='disabled'" for the input field, as well as displaying "*********..." instead of the real license. Just to make it clear that the code is set otherwere. The buttons to disable and activate a license should be obsolete as well. That is, because you cannot actually achieve anything with the input field, when a value is defined in wp-config (which is good!). So the UI should reflect that.

2. If the key is set in wp-config, you need to trigger the "Check license" initially in the code. Otherwise the license will never be activated. That is the point I haven't solved neither. But you might know the appropriate piece of code where it belongs to. Check should be like if license != active and defined(...) try to activate it.

That's from my side, thanks for the quick implementation :-)

Jan

pippinsplugins:
I'm not sure clicking `Activate` from the UI makes sense when the license key is entered in the wp-config.php file but I'm also not sure auto-activating it makes sense either. The main scenario I'm thinking of is when a license key has been entered in the wp-config.php file but is invalid for one reason or another. In that case, we'll still need to show some sort of feedback to the user. Currently that feedback is shown after clicking `Activate` or after the `check_license()` action has run.

We could attempt to the activate the license during the install routine if it is set in wp-config.php and then just fallback to the existing `check_license()` method for periodically validating it.

JanThiel:
[...]

PR #1268, opened 2016-05-06
issue/1229 – Make license key configurable via wp-config.php

DrewAPicture:
Issue: #1229 

Introduce a new constant, `AFFILIATEWP_LICENSE_KEY`, which can be defined in wp-config.php to override the license for all sites in a network or simply override the license key in a single site.

The new constant is checked in one place, a new static method in `Affiliate_WP_Settings`, `get_license_key()`.


pippinsplugins:
@DrewAPicture looks like we need to bring this / 1.9 up to date with master as well.

includes/admin/settings/class-settings.php
```diff
@@ -358,21 +371,14 @@ function get_registered_settings() {
...
-			$debug_mode_desc = __( 'Check this box to enable debug mode. This will turn on error logging for the referral process to help identify problems.', 'affiliate-wp' );
-		}
+		/**
+		 * Fires before attempting to retrieve registered settings.
+		 *
+		 * @since 1.9
+		 *
+		 * @param Affiliate_WP_Settings $this Settings instance.
+		 */
+		do_action( 'pre_get_registered_settings', $this );
```
pippinsplugins: Shouldn't this be prefixed with `affwp_` or is there a reason for using the same action as WP core?

DrewAPicture: @pippinsplugins Nope, that's a good call. I also thought about it a little bit, and if we don't want to open up an entry point for third-party developers with this hook, we could also just manually fire those two callbacks at the point of the hook. Thoughts?

pippinsplugins: I like having the hooks.

DrewAPicture: okie dokie!

Issue #1272, opened 2016-05-06
affwp_version option name isn't updated when updating AffiliateWP


sumobi:
The `affwp_version` option name (as far as I can see) does not update when a user updates their existing copy of AffiliateWP. 

It only updates when:

1. There's an upgrade routine (for v1.8 there wasn't one)
2. When AffiliateWP is installed for the first time
3. When AffiliateWP is deactivated, and then activated again

Because of this, customers will not see the checkboxes in Affiliate Area Tabs to remove the existing tabs, unless they deactivate and reactivate AffiliateWP.

I think we should update the `affwp_version` option name when a customer upgrades from one version to the next, regardless of if there is an upgrade routine.

pippinsplugins:
Good catch!

DrewAPicture:
Added a PR: #1294

DrewAPicture:
@pippinsplugins Added a new commit to the PR that uses `version_compare()` with `<>` as an operator. Also supplies an empty fallback value on the `get_option()` call. Waiting for tests to pass.

pippinsplugins:
Looks good to me, merge once tests pass please.

DrewAPicture:
Merged #1294

PR #1294, opened 2016-05-17
issue/1272 – Not all upgrades trigger upgrade routines

DrewAPicture:
Issue: #1272


includes/admin/class-upgrades.php
```diff
@@ -47,9 +47,15 @@ public function init() {
 
 		if ( version_compare( $version, '1.7.14', '<' ) ) {
 			$this->v1714_upgrades();
-		}	
+		}
+
+		$current_version = get_option( 'affwp_version' );
+
+		if ( $current_version !== AFFILIATEWP_VERSION ) {
```
pippinsplugins: Can we use `version_compare` here instead to make sure there are no potential type casting conflicts?

DrewAPicture: Yeah, I was thinking this would cover both upgrade and downgrade minus an
upgrade routine. So what if we do both greater- and less-than here with
`version_compare()`?

pippinsplugins: Sounds good

PR #1307, opened 2016-05-22
Pass $active_tab to affwp_affiliate_dashboard_top *_notices and *_bottom hooks

### LINKED ISSUE #1306: Add $active_tab to affwp_affiliate_dashboard_bottom hook
We should pass `$active_tab` to the `affwp_affiliate_dashboard_bottom` action hook so it makes it very easy to prevent the content of the new tab from showing on any other tab.

One _could_ simply use  `affwp_get_active_affiliate_area_tab()` in their code, but I would argue that since we already pass it to the `affwp_affiliate_dashboard_tabs` hook, and the `get_template_part` call we should do the same for `affwp_affiliate_dashboard_bottom`.

sumobi:
Fix #1306


sumobi:
👍

pippinsplugins:
Looks good.

PR #1314, opened 2016-05-25
issue/1317 - update malay translation

sumobi:
Props Moohamat Thoiruddin


michaelbeil:
This should be good to go.

Issue #1327, opened 2016-06-02
Test coverage for referral-functions.php


DrewAPicture:

DrewAPicture:
@pippinsplugins These are ready for review in #1328.

pippinsplugins:
Merged!

Issue #1301, opened 2016-05-21
Implement Core objects


DrewAPicture:
This is the integration/implementation for #1184.

DrewAPicture:
I think we're ready to go here. As discussed with Pippin on a walkthrough call last week this basically introduces:

* Dependable caching wrappers for the four core objects: affiliates, referrals, creatives, and visits: `AffWP_Affiliate`, `AffWP_Referral`, `AffWP_Creative`, and `AffWP_Visit` routed through their respective `affwp_get_*()` counterparts
* Introduces several new methods and capabilities to retrieving said objects or related properties, including a new `ID` property alias, e.g. no more typing `$referral->referral_id`, just `$referral->ID` :-p
* Puts us well on our way toward implementing REST endpoints (!)

Next steps: We should first merge #1184 to introduce the actual object code, and then this one to introduce the implementation.

pippinsplugins:
This is merged into `release/1.9`.

I'm going to close this but leave #1184 open for a bit longer in case any problems are discovered. Any further discussion needed can happen on #1184

PR #1328, opened 2016-06-03
issue/1327 – referral-functions.php tests

DrewAPicture:
Issue: #1327 

Adds test coverage for everything in referral-functions.php except for `affwp_calc_referral_amount()` and `affwp_count_referrals()`, which have other issues that should be addressed separately. Neither should be affected by changes in #1184 and #1301.


pippinsplugins:
Looks good!

PR #1338, opened 2016-06-05
issue/1343 - Add 'affwp_insert_pending_referral' filter to RCP integration

clickstudio:
Issue: #1343

This will bring the integration inline with all other integrations who use this filter.

Could also just use 'insert_pending_referral()' here which fires the filter ?????

I have not tested this.
I am developing an add-on that requires the filter and noticed it was missing for this integration.


pippinsplugins:
Looks good! We'll get it merged in!

clickstudio:
Thanks @pippinsplugins :-)

michaelbeil:
#1343

DrewAPicture:
@clickstudio Looks pretty good. Fixed a couple of notices in #1483 as a continuation of this PR. Closing in favor of that.

PR #1347, opened 2016-06-10
Use return in lieu of continue (PHP7 compat)

### LINKED ISSUE #1346: PHP7 continue must have a loop context
In `integrations/class-memberpress.phpL352`, the `continue;` must be removed to prevent a fatal error in  PHP7 due to context restriction changes.

ramiabraham:
Removes a continue used in the MemberPress integration. Fixes #1346


ramiabraham:
Closing in favor of pulling against 1.8.3.

Issue #1237, opened 2016-04-27
"No integration enabled" and "Enter/activate license key" admin notices are visible to logged-in affiliates


wordmaid:
If there are no integrations enabled, or the license key has not been entered/activated, a logged-in affiliate can see these admin notices on the WordPress admin dashboard. These notices should probably only be visible to admins.

<img width="1510" alt="screen shot 2016-04-27 at 11 49 57 am" src="https://cloud.githubusercontent.com/assets/11830108/14837735/0fa4a2aa-0c6f-11e6-8e4b-811080f59198.png">

DrewAPicture:
Added a PR: #1350

michaelbeil:
PR looks good!

pippinsplugins:
@DrewAPicture PR looks good minus one minor detail. The `@since` tag has the wrong version on it.

DrewAPicture:
@pippinsplugins The code in question was added on the 0.2 tag, which is where that came from. If you have a different version you'd like me to use, I'm all eyes ;)

pippinsplugins:
Ah! I forgot we had taged pre 1.0 versions.

Let's use 1.0 please.

DrewAPicture:
Tests passed. Merged #1350.

PR #1350, opened 2016-06-11
issue/1237 – Bail early on displaying admin notices if the current user can't manage affiliates

DrewAPicture:
Issue: #1237


includes/admin/class-notices.php
```diff
@@ -1,15 +1,34 @@
 <?php
-
+/**
+ * AffiliateWP Admin Notices class
+ *
+ * @since 0.2
```
pippinsplugins: This version isn't correct. We've never had a `0.2` version.

PR #1374, opened 2016-06-25
issue/872 – Display visit date and time in the Visits tab in the affiliate area.

DrewAPicture:
Issue: 872


templates/dashboard-tab-visits.php
```diff
@@ -45,6 +46,9 @@
 								<i></i>
 							</span>
 						</td>
+						<td>
+							<?php echo esc_html( mysql2date( 'Y-m-d H:i:s', $visit->date ) ); ?>
```
pippinsplugins: Shouldn't this use date_i18n() instead?

pippinsplugins: Also probably best to set the format with the `date_format` option.

DrewAPicture: @pippinsplugins I thought about that, except we also want to show then time, at which point we're getting into *two* options pulls and logic to display it, which I'm not sure necessarily belongs in a template. So we can create a helper or something, but in this instance, I used the same hard-coded date format used in the list table.

pippinsplugins: Fair point. Let's leave as is for now then.

DrewAPicture: @pippinsplugins In response to the question about using `date_i18n()`, we should probably pass `true` as the third argument to `mysql2date()` to ensure the time string is passed through `date_i18n()` and therefore translated. The reason `date_i18n()` isn't used here directly, is that it expects a unix timestamp whereas `mysql2date()` expects a date string which is then converted into a unix timestamp.

PR #1440, opened 2016-07-15
issue/1184 – Enable lazy loading an affiliate\'s user object with magic $user property

DrewAPicture:
Issue: #1184


pippinsplugins:
Looks good. Let's merge once tests pass.

Issue #1470, opened 2016-07-25
Visits/cookies not being stored with fallback tracking/pretty URLs/campaigns


sumobi:
To replicate:

- Master branch
- Enable "Use Fallback Referral Tracking Method?" option in Settings -> Misc

## Issue 1: 

PR: #1472 

When fallback tracking is enabled & pretty affiliate URLs are used, referral links generated within the Affiliate Area that include a "campaign" do not work. They don't store any cookies, which means no visit stored.

The pretty URLs generated in the Affiliate Area when using a campaign look like this:

http://mysite.com/ref/john?campaign=test

However, this URL works (note the forward slash after john):

http://mysite.com/ref/john/?campaign=test

With fallback tracking **disabled**, both variations work as expected:

http://mysite.com/ref/john?campaign=test
http://mysite.com/ref/john/?campaign=test

Clue:

If you turn on debugging you'll see this error log message when using the URL above that doesn't store a visit:

> No user account found for given affiliate ID or login during get_fallback_affiliate_id()

https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/class-tracking.php#L440

## Issue 2 (related)

When a custom slug is used (via the Custom Affiliate Slugs add-on) with a pretty URL, and with fallback tracking enabled, visits are not being stored. 

E.g. http://mysite.com/ref/mycustomslug

Same log message:

> No user account found for given affiliate ID or login during get_fallback_affiliate_id()

I suspect it's related to the same issue. Without a pretty URL, it stores a visit + cookies fine.

E.g. http://mysite.com/?ref=mycustomslug

@pippinsplugins @DrewAPicture Can we please give this some attention as soon as possible?

pippinsplugins:
`http://mysite.com/ref/john?campaign=test` fails because we split the URL on `/`:
```
var path = window.location.pathname.split( '/' );
$.each( path, function( key, value ) {
	if( AFFWP.referral_var == value ) {
		ref = path[ key + 1 ];
	}
});
```
Since there is no `/` after the referral var, `ref` is never found.

sumobi:
👍

DrewAPicture:
Added a PR for the first part: #1472. I think since we need to be able to reliably split on `/`, the best option is to just ensure that a trailing slash is present after the referral var value in generated links.

sumobi:
@DrewAPicture 👍 

[...]

PR #1472, opened 2016-07-26
issue/1470 – Ensure a trailing slash always falls after the reverral var value.

DrewAPicture:
Issue: #1470


pippinsplugins:
Updated this to go into `release/1.9`.

PR #1480, opened 2016-07-28
#issue/1452: Add referrals export cap.

ramiabraham:
This introduces the export_referral_data capability, allowing arbitrary users to export referral data when this cap is mapped to them.

This also adds the new cap to the AffiliateWP default administrator roles.

+ Related unit test assertion.

Fixes #issue/1452.


ramiabraham:
Here's a quick plugin to test this; change user id array to any non-admin user:

https://gist.github.com/ramiabraham/5ca12e9f3098df053535e1e6bd96c3d9

PR #1496, opened 2016-08-02
Issue/1466 – Add website to Edit Affiliate

### LINKED ISSUE #1466: Allow site admins to see affiliate's submitted "Website URL" in Edit Affiliate screen
PR: #1496 

I'm going through our own affiliates and there are a couple of things I've noticed regarding affiliate registration and the affiliate's submitted website URL:

1. The affiliate's website URL is not visible to the admin anywhere. I'm going through each affiliate and searching for them/their business, or using their email addresses to (hopefully) find their site where they have promoted/will promote our product. I know this is saved in the database somewhere, but it would be incredibly useful if this URL was visible on the affiliate's Edit screen. It could save admins a significant amount of time searching for affiliates' websites. This should be an editable field in case any spelling mistakes were made during the initial submission, or in case the URL where the site is promoted changes in the future. The website URL should also be able to be edited by the affiliate from the Affiliate Area, which then updates in the admin (again, in case it is correct or it changes).
2. Similarly, the website URL should be able to be entered when manually adding a new affiliate from the Add New Affiliate screen. If a site owner were to approach potential affiliates, or for some reason a new affiliate doesn't register through the registration form (they meet them offline for example), there is nowhere to capture where they will be promoting the product/site online.

It would allow site owners to be able to easily find their affiliates' websites. What do people think?

DrewAPicture:
Fixes #1466


includes/actions.php
```diff
@@ -41,6 +40,17 @@ function affwp_remove_query_args( $query_args ) {
 		'_wpnonce'
 	);
 
+	$edit_actions = array(
```
pippinsplugins: Pretty sure this needs to have the `add_` version too, yes?

DrewAPicture: @pippinsplugins Is refreshing on the `add_` context a common thing? I honestly don't know. I'd have to look at the context in which the action is leveraged.

pippinsplugins: Oh right! Ignore me :D

Issue #471, opened 2014-12-05
Payouts log


pippinsplugins:
* [x] PR: #1501
* [x] WP-CLI PR: #1539 

I'd really like to see a payouts log system created that keeps track of all payouts for affiliates. Each time an affiliate receives a payout, it would log it. This would need to be usable with automatic payouts and manual payouts.

This would allow affiliates to see a history of their payments as well.

codelion7:
@pippinsplugins would this entail creating a class and a db functions file for storing and displaying the data? We would probably need an admin interface and a tab in the affiliate area as well.

pippinsplugins:
Yes, it'd be similar to our other database tables / UIs

pippinsplugins:
Thinking about this a bit, I think our database table should contain the following columns:
- payout_id
- affiliate_id
- amount
- referral_ids
- date
- payout_method

Each time a payout is processed, an entry will need to be recorded for each affiliate account separately. For example, if a payout is processed and two affiliate's receive a payment, two payout log entries will be recorded.

We can introduce a new list table UI, much like the referrals table, that displays the complete history of payouts. Payouts should also be made available inside of the affiliate area for affiliates to see.

Along with the database class, I propose that we introduce an `AffWP_Payout` object at the same time.

codelion7:
That sounds perfect @pippinsplugins. The object would allow for further integration and extensibility.

pippinsplugins:
Modified DB structure proposal:

**Payout logs DB:**
- payout_id
- affiliate_id
- amount
- date
- payout_method

**New columns on referrals DB**
- payout_id

DrewAPicture:
Started a PR: #1501 

There's a TODO list in the PR description. Feel free to add anything I missed.

pippinsplugins:
TODO looks good.

DrewAPicture:
**Tuesday progress:**
* Built TODO list
* Stubbed out `AffWP\Affiliate\Payout`

**Wednesday progress:**
* Stubbed out `Affiliate_WP_Payouts_DB`
* Set up compat w/ `AffWP\Object`
* Stubbed out `payout-functions.php` + basic tests for `affwp_get_payout()`
* Got the first round of feedback from @pippinsplugins and @ramiabraham

DrewAPicture:
**Thursday Progress:**

* Added `status` column and renamed `referral_ids` column to `referrals`
[...]

PR #1501, opened 2016-08-03
issue/471 – Payout logs

DrewAPicture:
Issue: #471 

TODO:

* [x] `AffWP\Payout` object
* [x] `Affiliate_WP_Payouts_DB` class
* [x] Convert `AffWP\Object` to support db subgroups
* [x] `includes/payout-functions.php`, – added `affwp_get_payout()`, `affwp_add_payout()`, `affwp_delete_payout()`, `affwp_get_payout_referrals()`, and `affwp_get_payout_status_label()`
* [x] `includes/payouts/*` files
* [x] Convert includes/payouts/edit.php to view.php
* [x] `payout_id` column for referrals DB
* [x] Upgrade routine bits (left the trigger out for now)
* [x] Affiliate Area tab template
* [x] Integration with payout CSV
* [ ] WP-CLI commands
* [ ] REST endpoint(s)
* [x] Payouts DB tests, – Added tests for `add()`, `payout_exists()`, `get_column_defaults()`, `get_columns()`, `get_object()`, and `get_referral_ids()`. Tests for `get_payouts()` cover current functionality (need tests for amount )
* [x] Payout functions tests, – Full coverage
* [x] Payout object tests, – stubbed
* [x] Consolidate logic in `affwp_add_payout()` with `add()` and move it

Probably other stuff I can't think of just now. Full court press.


includes/admin/referrals/referrals.php
```diff
@@ -519,6 +519,14 @@ public function process_bulk_action() {
 			}
 
 			if ( 'mark_as_paid' === $this->current_action() ) {
+				if ( $referral = affwp_get_referral( $id ) ) {
+					affwp_add_payout( array(
+						'affiliate_id'  => $referral->affiliate_id,
+						'referrals'     => $id,
```
pippinsplugins: I think this should also pass the referral amount since we want the amounts stored in payouts.

pippinsplugins: Ah, I see where it pulls it from the referrals below. Ignore this.

includes/payout-functions.php
```diff
@@ -28,6 +28,129 @@ function affwp_get_payout( $payout = 0 ) {
...
+		}
+	}
+
+	if ( empty( $data['affiliate_id'] ) ) {
+		return false;
+	} else {
+		$args['affiliate_id'] = absint( $data['affiliate_id'] );
+	}
+
+	if ( ! empty( $data['amount'] ) ) {
```
pippinsplugins: This is smart and I like it, though I do have a concern that there may be a scenario where we want a payout stored with a 0 amount. Passing `0.00` would work and maybe that's sufficient, but I think we should consider whether `0` and/or `false` should also be supported.

Issue #668, opened 2015-04-24
Remove redirect_canonical on WP 4.3+


pippinsplugins:
Thanks for https://core.trac.wordpress.org/ticket/25143, we can remove the madness that is:

https://github.com/affiliatewp/AffiliateWP/blob/master/includes/class-tracking.php#L43
https://github.com/affiliatewp/AffiliateWP/blob/master/includes/class-tracking.php#L44

sumobi:
Woot!

DrewAPicture:
Adding this to 1.9 since we're bumping to 4.4 minimum.

pippinsplugins:
PR: https://github.com/AffiliateWP/AffiliateWP/pull/1506

pippinsplugins:
@sumobi could you test this please?

sumobi:
@pippinsplugins Anything special I need to look out for? I've been testing referral links on the homepage (static page and default) and all seems to work fine.

sumobi:
Actually, I'm having trouble generating cookies with using pretty affiliate URLs on the homepage (when static page is set), non-pretty are working fine.

sumobi:
Doesn't seem to be related to this PR, broken for me in `master` also.

sumobi:
Broken with public 1.8.7 also, hmm.

sumobi:
Static page doesn't seem to matter, broken with default also. Either it's very busted or I'm just doing it wrong.

sumobi:
Something definitely weird going on here. After making sure my permalinks are flushed etc there seems to be an issue when using a static page set to the homepage and pretty affiliate links. The `/ref/10` is being stripped from the URL and no cookies are loaded. I was seeing sporadic behaviour between browsers but now it's happening in firefox, chrome, safari. I originally thought it was master as well but I'm fairly certain some caching was involved.

pippinsplugins:
@sumobi which integrations are enabled?

michaelbeil:
There is a customer that is experiencing this but I cannot replicate it locally. 

What version of WP @sumobi?

sumobi:
@pippinsplugins tried with none, WooCommerce, and only EDD

@michaelbeil 4.6

Have also replicated the issue on my other machine with a static front page set as above. Master is fine, but switching to this branch instantly strips off the `/ref/1` from the URL and no cookies stored.

pippinsplugins:
Ugh.

pippinsplugins:
[...]

PR #1506, opened 2016-08-04
Set $query_var to true when registering rewrite - Issue/668

pippinsplugins:
Fixes: #668

Also removes redirect_canonical

### DIFF
--- a/includes/class-rewrites.php
+++ b/includes/class-rewrites.php
@@ -34,7 +34,6 @@ class Affiliate_WP_Rewrites {
 
 		}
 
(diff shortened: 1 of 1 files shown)


michaelbeil:
Restarted Travis and this passes now.

includes/class-rewrites.php
```diff
@@ -132,31 +131,4 @@ public function unset_query_arg( $query ) {
 
 	}
```
DrewAPicture: We should properly deprecate this method in case somebody was using it.

pippinsplugins: @DrewAPicture how's that?

PR #1529, opened 2016-08-10
issue/1008 - Unhide affiliate registration fields

### LINKED ISSUE #1008: Keep username and email fields but set them to disabled for logged in users
A common support ticket we get is "I can't see the email or username fields on my affiliate registration form". This is obviously because the user is already logged in.

I'd love to hear everyone's thoughts on what we can do to make this clearer. We could for instance keep the fields there, set them to disabled, and show the user's current username and email address, much like we do on the edit affiliate screen for the user ID.

Since the fields are already there for a logged out user, it wouldn't hurt to keep them there for a logged-in user and would prevent many support tickets for AffiliateWP, and the Affiliate Forms for X add-ons which mimic the same behaviour.

sumobi:
Fix #1008

### DIFF
--- a/includes/class-register.php
+++ b/includes/class-register.php
@@ -111,7 +111,7 @@ class Affiliate_WP_Register {
 			// Loop through required fields and show error message
 			foreach ( $this->required_fields() as $field_name => $value ) {
 
-				if( ! empty( $value['logged_out'] ) ) {
+				if ( ! empty( $value['logged_out'] ) ) {
 					continue;
 				}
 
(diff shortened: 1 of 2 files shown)


includes/class-register.php
```diff
@@ -228,19 +228,22 @@ private function register_user() {
 
 		$status = affiliate_wp()->settings->get( 'require_approval' ) ? 'pending' : 'active';
 
-		if ( ! is_user_logged_in() ) {
-
+		if ( $_POST['affwp_user_name'] ) {
```
pippinsplugins: If username is not posted, this will cause an undefined index notice. Using `! empty( $_POST['affwp_user_name'] )` will work properly without throwing a notice.

sumobi: Yeah, whoops 👍

PR #1585, opened 2016-09-02
issue/1543 - Properly support thousand and decimal separators in affwp_sanitize_amount()

### LINKED ISSUE #1543: Amount Sanitize function not correct
https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/misc-functions.php#L109

The regex is incorrect and removes any commas, which then makes most of the logic after it useless ;)

Not sure why, but this function is different than EDD's version (which works correctly)

https://github.com/easydigitaldownloads/easy-digital-downloads/blob/master/includes/formatting.php#L24

pippinsplugins:
Pull in improvements made to `edd_sanitize_amount()` to properly support thousand and decimal separators. Resolves #1543


ramiabraham:
Looks good here:

- Thousand and decimal separators displaying correctly. 
    USD:
    amt
    JPY:
    amt

- Checks for non-negatives working as intended as well.

PR #1587, opened 2016-09-02
Add plugin compatibility for Simple Page Test. Resolves #1586

### LINKED ISSUE #1586: Adding referral var to static home page with Simple Page Tester plugin active causes blog posts to load
When the [Simple Page Tester](https://wordpress.org/plugins/simple-page-tester/) plugin is activated, adding a referral variable to the home page (set to a static page) causes the blog index to be loaded instead.

pippinsplugins:
#1586


pippinsplugins:
Redoing PR.

Issue #1670, opened 2016-09-30
Styles cannot be dequeued on the [affiliate_login] shortcode


sumobi:
The standalone login form which uses the `[affiliate_login]` shortcode outputs its CSS file directly inside the `affiliate_login()` method.  This isn't ideal for 2 reasons:
1. It loads the `forms.css` file just before the closing `</body>` of the page. CSS files should generally be loaded in the `<head>` 
2. The `forms.css` file cannot be dequeued so a theme can't use its own CSS styling for the form.

Like we did for https://github.com/AffiliateWP/AffiliateWP/issues/1026 I propose we remove the `wp_enqueue_style` function call and load the stylesheet inside our `affwp_frontend_scripts_and_styles()` function since the CSS has already been registered. 

This will allow a theme to dequeue the CSS file, as well as the forms.css file being correctly loaded in the `<head>`.

DrewAPicture:
See the PR on #1672, which (indirectly) addresses this.

DrewAPicture:
After some discussion with @sumobi the other day in Slack, our best option to make scripts or styles dequeueable might be something along these lines:
```php
function affwp_enqueue_style( $handle, $context = '' ) {
    /**
     * Filter whether to enqueue the stylesheet.
     * 
     * The dynamic portion of the hook name, `$handle` refers to the stylesheet handle.
     * 
     * @since 1.9.4
     * 
     * @param bool   $enqueue Whether to enqueue the stylesheet. Default true.
     * @param string $context Context under which to enqueue the stylesheet.
     */
    if ( true === apply_filters( "affwp_enqueue_style_{$handle}", true, $context ) ) {
        wp_enqueue_style( $handle );
    }
}
```

 ```php
add_filter( 'affwp_enqueue_style_affwp-forms', function( $enqueue, $context ) {
    if ( 'login' === $context' ) {
        $enqueue = false;
    }
    return $enqueue;
} );
```

pippinsplugins:
I like that proposal.

DrewAPicture:
OK. I'll work on a PR.

DrewAPicture:
I've added a secondary PR with the `affwp_enqueue_*()` approach. It avoids using `has_shortcode()` in a couple of places and simultaneously introduces the ability to selectively disable enqueuing within individual shortcode context. Please test :-)

pippinsplugins:
I really like this approach.

sumobi:
[...]

PR #1671, opened 2016-09-30
issue/1670 - Remove wp_enqueue_style()

sumobi:
+ add wp_enqueue_style inside affwp_frontend_scripts_and_styles()

#1670


includes/scripts.php
```diff
@@ -117,6 +117,10 @@ function affwp_frontend_scripts_and_styles() {
 	$suffix = ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) ? '' : '.min';
 	wp_register_style( 'affwp-forms', AFFILIATEWP_PLUGIN_URL . 'assets/css/forms' . $suffix . '.css', AFFILIATEWP_VERSION );
 
+	if ( has_shortcode( $post->post_content, 'affiliate_login' ) ) {
+		wp_enqueue_style( 'affwp-forms' );
+	}
+
```
sumobi: This seems like the best way to do it since the [affiliate_area] and [affiliate_registration] shortcodes need everything else (dashicons, recaptcha, frontend.js) but the login shortcode does not.

DrewAPicture: As mentioned in [#1672](https://github.com/AffiliateWP/AffiliateWP/issues/1672#issuecomment-250994890) we should avoid using `has_shortcode()` at run-time at all costs. It's extremely non-performant function.

includes/scripts.php
```diff
@@ -117,6 +117,10 @@ function affwp_frontend_scripts_and_styles() {
...
 	if ( has_shortcode( $post->post_content, 'affiliate_area' ) || has_shortcode( $post->post_content, 'affiliate_registration' ) || apply_filters( 'affwp_force_frontend_scripts', false ) ) {
```
michaelbeil: We should probably rework using `has_shortcode()` here as well.

DrewAPicture:
Accidentally pushed to the wrong branch, but this PR is being replaced with #1748  anyway. Closing.

PR #1679, opened 2016-10-02
added action hooks to dashboard-tab-payouts template

flowdee:
Would be great to be able to hook in here as well as in the other templates


pippinsplugins:
Thanks @flowdee, we'll look at including this (or some version of it) in 1.9.3.

flowdee:
Yep great! My new addon is waiting for this ;)

templates/dashboard-tab-payouts.php
```diff
@@ -18,13 +18,16 @@
 	);
 	?>
 
+	<?php do_action( 'affwp_payouts_dashboard_before_table', affwp_get_affiliate_id() ); ?>
```
DrewAPicture: We've started ensuring all new hooks get documented when they're added. For an example on the formatting, see here: https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/affiliate-functions.php#L1145

Note, the hook DocBlock must _directly precede_ the hook line for our parser to pick it up. You can use `@since 1.9.4`.

flowdee:
Thanks for the review @DrewAPicture. Added the documentation as wished.

DrewAPicture:
Looks good to me. Thanks for the adding the docs!

PR #1680, opened 2016-10-02
issue/1685 – Fixed affwp_payout_row_actions filter

flowdee:
Issue: #1685

`affwp_affiliate_row_actions` should be `affwp_payout_row_actions`


michaelbeil:
Good catch!

PR #1716, opened 2016-10-14
Adding action affwp_add_affiliate

davefx:
Affiliates get added from several places. Many of them doesn't involve the launch of any action. I've added the action affwp_add_affiliate so we can have a single action to modify or add data to the just-inserted affiliate.


davefx:
Sorry. Just discovered affwp_insert_affiliate function hook.

PR #1723, opened 2016-10-18
issue/1817 – Add an affwp_woocommerce_get_products_line filter

davefx:
Issue: #1817

Adding filter affwp_woocommerce_get_products_line to allow the customization of Woocommerce product line


pippinsplugins:
@davefx Could you provide me with an example of when this is needed?

davefx:
We are working in a plugin that is able to apply different variable rates depending on several factors, and we are using this filter to add extra information about the applied rate and the reason of its application.
Using this, we can generate detailed reports of the applied rates.

pippinsplugins:
:+1:

We'll get this is 1.9.5.

DrewAPicture:
@davefx In the future, all new hooks should have accompanying DocBlocks. I'll take care of this one though :)

PR #1724, opened 2016-10-18
Adding new filters in export class

davefx:
Adding new filters affwp_export_get_data_csv_cols and affwp_export_get_data_line to extend the data available for CSV reports and for the filters affwp_export_get_data and affwp_export_get_data_<export_type>.

This is needed, for example, to include data from $referral->custom field in the exported reports.


pippinsplugins:
@davefx there are already filters for these. See the base export class: 

https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/admin/tools/export/class-export.php#L84

davefx:
Yes, we have for the cols, but we don't have any to add new information line-by-line to the passed data without launching database queries.
I'll create again a PR with this last one, that is needed if we want to add $referral->custom to the data provided to the filters.

pippinsplugins:
You can loop through the data, though per line could be better.

We can re-consider this in 1.9.5. Thanks!

PR #1725, opened 2016-10-18
issue/1818 – Add a 'affwp_export_data_line' filter

davefx:
Issue: #1818

Adding new filter 'affwp_export_data_line' so we can provided extra information from the referrals to later filters without launching extra queries.

We are using this to pass $referral->custom to the filters (without this, to access $referral->custom we should make an extra query).


DrewAPicture:
@davefx In the future, all new hooks should have accompanying DocBlocks. I'll take care of this one though :)

Issue #1641, opened 2016-09-26
Lazy load reports


pippinsplugins:

pippinsplugins:
PR with one approach to improving this: https://github.com/AffiliateWP/AffiliateWP/pull/1739

PR #1739, opened 2016-10-21
Fix/Reduce Reports Memory Footprint

davefx:
Until now, data for all report graphs is loaded just by loading the displaying the reports tabs.
With this change, the graphs data won't be loaded until just before displaying them, hugely reducing the  memory footprint.


pippinsplugins:
Thanks @davefx! This is along the lines of what we have planned for lazy loading the reports: https://github.com/AffiliateWP/AffiliateWP/issues/1641

We'll look at it during the 2.0 development cycle.

Issue #1752, opened 2016-10-25
Rejected referrals to be updated to Unpaid if associated order is completed


pippinsplugins:
If a referral gets marked as Rejected due to a failed order and then the customer comes back and successfully completes the order, the referral remains Rejected, instead of changing to Unpaid.

I don't think this is right. The referral should change to `Unpaid`, but only so long as the current status is `Pending` or `Rejected`. A `Paid` referral should remain unchanged.

pippinsplugins:
@sumobi could you test this please?

DrewAPicture:
PR looks good.

sumobi:
@pippinsplugins What's the best way to test this with EDD? 

I tested with EDD and Stripe payment gateway, using a declined credit card number to simulate a declined card, but EDD doesn't create the order initially, unless it's successful.

I then tried a normal Stripe payment to create the order, then set it to "revoked" from the payment history (affwp setting is enabled to revoke referral), but that didn't reject the referral. As per the code I can see that the referral is only revoked when refunded, or the payment is deleted entirely (but then I can't approve it again)

So then I ran a test payment and refunded the order which rejected the referral. I was thinking I could then "complete" the payment and the rejected referral would move to "unpaid" again (as per this PR) but that didn't happen.

Side discussion, I think we haven't accounted for some scenarios when a payment is manually modified from the admin, where the referral should be rejected, i.e revoking a payment in EDD, setting it to "failed", "abandoned", "cancelled" etc.

pippinsplugins:
WooCommerce would be the easiest. It's the only integration I'm currently aware of where this happens.

1. Setup PayPal standard gateway
2. Go to checkout
3. Submit and go to PayPal
4. Click cancel on PayPal
5. Return to checkout and change payment method to something else
6. Complete order

Referral should end up as `Unpaid`.

davefx:
I think that the actual problem is that we are actually marking a referral as rejected when its corresponding WooCommerce order gets into 'failed' status. 
An WooCommerce in a failed status is not a revoked nor a refunded order, so I think that the problem is generated because we are rejecting something that shouldn't be rejected at the first place. I think this referrals should remain in 'pending' status.

michaelbeil:
Hmm, that is an interesting thought to remove that action @davefx.

Perhaps we can merge both PRs as a part of this. Well, after further thought, the above PR from Pippin should account for the WC order fail.

We should also address @sumobi's comment further:

[...]

PR #1757, opened 2016-10-26
Don't revoke referral when Woo order fails #1752

davefx:


davefx:
This would also fix #1752

Issue #1763, opened 2016-10-27
DRY up outputting navigation tabs


DrewAPicture:
As [pointed out](https://scrutinizer-ci.com/g/AffiliateWP/AffiliateWP/indices/1490116/duplications/942941) by Scrutinizer, we have essentially duplicated the same nav-tab output markup four different places in AffWP core. We should DRY this up.

> _In software engineering, don't repeat yourself (DRY) is a principle of software development, aimed at reducing repetition of information of all kinds, especially useful in multi-tier architectures._

DrewAPicture:
Added a PR: #1764 

Introduces `affwp_navigation_tabs()` and implements it in all 4 places.

Please test that the tabs are output as expected on:
1. Affiliates > Reports
2. Affiliates > Tools
3. Affiliates > Settings
4. Affiliates > Add-ons

pippinsplugins:
Some minor changes requested on PR.

DrewAPicture:
@pippinsplugins I've updated the PR with changes addressing your feedback.

pippinsplugins:
Looks and works great, merged!

PR #1764, opened 2016-10-27
issue/1763 – Introduce affwp_navigation_tabs() to output nav tabs in various AffWP screens

DrewAPicture:
Issue: #1763


includes/misc-functions.php
```diff
@@ -752,4 +752,27 @@ function affwp_get_current_screen() {
...
+ * Outputs navigation tabs markup in core screens.
+ *
+ * @since 1.9.5
+ *
+ * @param array  $tabs       Navigation tabs.
+ * @param string $active_tab Active tab slug.
+ * @param array  $query_args Optional. Query arguments used to build the tab URLs. Default empty array.
+ */
+function affwp_navigation_tabs( $tabs, $active_tab, $query_args = array() ) {
+	foreach ( $tabs as $tab_id => $tab_name ) {
```
pippinsplugins: Let's do a safety check here to ensure `$tabs` is an array. Also, let's add a hook before and after the tabs, and maybe even between each tab.

PR #1772, opened 2016-10-31
Issue/1773 – Wrong handling of time offset in reports

davefx:
As date expects unix time in GMT, now current_time call is made with $gmt=1.
If not, in Australia time the last day of the month from noon to midnight we are getting empty reports for the following month


davefx:
Fixes #1773

michaelbeil:
Are you saying that the last twelve hours of the final day of any month when included in a report will show an empty graph and quick stats on the reports screen when in Australian time?

What happens when the last day is excluded?

davefx:
Only if "this month" is selected. That's exactly how I've realized today.

davefx:
The bug also depends on the PHP timezone configuration. It only shows if configured PHP default timezone isn't UTC.

includes/class-graph.php
```diff
@@ -105,6 +112,53 @@ public function __construct( $_data = array() ) {
 
 	}
 
+
+
+
+	/**
```
DrewAPicture: These test methods should exist only within the scope of the test suite. This is because we need to be able to rely on (WordPress) core's default `current_time()` behavior when we call `current_time( 'timestamp', true )`. WordPress core _explicitly_ defaults to the UTC timezone when it loads, and then all core time/date functions manually generate GMT-based times off of that using the `gmt_offset` option.

So if we want to test for specific dates, such as the last day of the month, we can do that, but we should only do it within the confines of the test suite.

includes/class-graph.php
```diff
@@ -452,7 +506,7 @@ function graph_controls() {
 function affwp_get_report_dates() {
 	$dates = array();
 
-	$current_time = current_time( 'timestamp' );
+	$current_time = Affiliate_WP_Graph::current_time( 'timestamp', true );
```
DrewAPicture: See the adjacent PR I created for including an optional parameter to selectively override the timestime derived from `current_time()` with `$gmt = true`. This should work for the purposes of testing.

tests/test-graph.php
```diff
@@ -0,0 +1,160 @@
...
+		$this->assertEquals( $dates['year_end'], 2016 );
+
+	}
+
+	/**
+	 * @covers affwp_get_report_dates()
+	 */
+	public function test_get_report_dates_correct_this_month_at_the_end_of_the_month_utc_at_beginning_nz() {
+
+		date_default_timezone_set( 'Pacific/Auckland' );
```
DrewAPicture: We need to not override the timezone and instead rely on overriding/retrieving the `gmt_offset` option. This goes for all of the test methods.

davefx:
Ok. I've just made the changes proposed.
About the `date_default_timezone_set` call, this error gets reproduced when the theme has set a different timezone. I know that this is not a correct usage, but unfortunately some people have themes with these wrong settings, and with a lot of code depending on this setting too... With the proposed change, AffiliateWP would work correctly also with these *incorrect* scenarios.

DrewAPicture:
The main issue I have with venturing into correcting other developers `_doing_it_wrong()` is that we get pretty close to bridging into unit testing core. That aside, you kind have to draw the line somewhere. I recognize that fixing for the secondary case of bad-practice is appealing, but I feel like it ultimately raises our maintenance burden.

includes/class-graph.php
```diff
@@ -449,10 +449,12 @@ function graph_controls() {
  * @since 1.0
  * @return array
```
DrewAPicture: Let's get the DocBlock changes brought over from https://github.com/AffiliateWP/AffiliateWP/compare/master...davefx-patch-7 for the addition of the optional parameter.

tests/test-graph.php
```diff
@@ -0,0 +1,214 @@
...
+
+	}
+
+
+	/**
+	 * @covers affwp_get_report_dates()
+	 */
+	public function test_get_report_dates_correct_this_month_at_the_end_of_the_month_utc() {
+
+		date_default_timezone_set( 'UTC' );
```
DrewAPicture: And let's leave the timezone changes in place. We can always rip them out later they cause problems.

pippinsplugins:
Going to merge in #1774

Issue #1792, opened 2016-11-06
www.domain.com or domain.com not accepted as a valid URL


sumobi:
You can't use `www.domain.com` or even `domain.com` as a valid URL on the affiliate registration form (tested in firefox and chrome)

You will see:

> Please enter a valid website URL

I've seen 3 or more tickets on this so far. This is significant barrier for those signing up to become affiliates, and most won't know to add the domain protocol (e.g.http://). WP core allows both `www.domain.com` and `domain.com` as valid URLs when editing a user profile so we should also allow this.

sumobi:
Ready for testing

pippinsplugins:
See feedback on PR.

pippinsplugins:
Ok I'm happy with PR.

sumobi:
@michaelbeil want to give it a test?

DrewAPicture:
Looks good to me.

PR #1793, opened 2016-11-06
Issue/1792 - Improve URL validation

sumobi:
#1792


includes/class-register.php
```diff
@@ -69,9 +69,10 @@ public function process_registration( $data ) {
 					$this->add_error( $value['error_id'], $value['error_message'] );
 				}
 
-				if( 'affwp_user_url' === $field_name && false === filter_var( $field, FILTER_VALIDATE_URL ) ) {
-					$this->add_error( 'invalud_url', __( 'Please enter a valid website URL', 'affiliate-wp' ) );
+				if ( 'affwp_user_url' === $field_name && false === filter_var( esc_url( $field ), FILTER_VALIDATE_URL ) ) {
```
pippinsplugins: This shouldn't be escaped. We escape on output, not input.

includes/class-register.php
```diff
@@ -271,7 +272,7 @@ private function register_user() {
 		}
 
 		// website URL
-		$website_url = isset( $_POST['affwp_user_url'] ) ? sanitize_text_field( $_POST['affwp_user_url'] ) : '';
+		$website_url = isset( $_POST['affwp_user_url'] ) ? esc_url( $_POST['affwp_user_url'] ) : '';
```
pippinsplugins: Use `sanitize_text_field` instead

templates/register.php
```diff
@@ -23,7 +23,7 @@
 		$payment_email = sanitize_text_field( $_POST['affwp_payment_email'] );
 	}
 
-	$url    = sanitize_text_field( $_POST['affwp_user_url'] );
+	$url    = esc_url( $_POST['affwp_user_url'] );
```
pippinsplugins: `sanitize_text_field` was correct here.

includes/class-register.php
sumobi: `FILTER_VALIDATE_URL` requires a domain that includes the domain protocol, or it will return false. The reason why `domain.com` or `www.domain.com` are not being accepted as correct domains is primarily this reason, they're being passed without a protocol. On the WordPress edit profile screen, there are no validation errors. If you enter `domain.com`, WordPress assumes you meant to enter a "valid" domain and adds the domain protocol after save. Any other suggestions? `esc_url_raw()`?

Codex:

> Always use esc_url when sanitizing URLs (in text nodes, attribute nodes or anywhere else). 

WordPress itself uses esc_url in hidden input fields, form actions etc when it needs to pass the URL around.

templates/register.php
sumobi: `esc_url` is used so if there are validation errors, the page reloads and the URL entered into the URL field shows with a domain protocol (on output). Like the WordPress profile page, if you enter `domain.com` as the URL, the result when saving the page is the domain with the protocol, `http://domain.com`.

If we do use `sanitize_text_field`, the form will reload and the URL will **not** have the domain protocol. I think we should assume (like the edit profile page) that the user meant to enter a valid domain and educate them if there is an error by showing them what the domain should have been. If they wanted to enter `https` instead, they will also be able to correct it before resubmitting the form.

I can revert it back but a domain is not valid without a domain protocol so I think we'll be doing a service by educating the user.

includes/class-register.php
pippinsplugins: Hmm okay, I guess taht's fine then :)

Issue #1453, opened 2016-07-20
Add inline docs for all actions and filters


ramiabraham:
In discussion with @DrewAPicture, we were thinking it'd be wonderful to get inline docs defined for every hook in core.

PR #1833, opened 2016-11-22
issue/1453: Inline docs for all core actions.

ramiabraham:
- Provide inline documentation for all core actions.
- Filters in progress.

#1453


pippinsplugins:
❤️

Issue #1834, opened 2016-11-23
Add support for user-defined locales


ramiabraham:
WordPress 4.7 introduces `get_user_locale` ([changeset](https://core.trac.wordpress.org/changeset/38705)), which falls back to the standard `get_locale`.

Let's update our `load_textdomain()` loader method ([link to proposed affected line](https://github.com/AffiliateWP/AffiliateWP/blob/master/affiliate-wp.php#L504)) to support user locales.

ramiabraham:
Added a PR: https://github.com/AffiliateWP/AffiliateWP/pull/1835

pippinsplugins:
Looks good!

michaelbeil:
Celebrate good times now.

ramiabraham:
Cool cool, merged.

ramiabraham:
New PR with small adjustment: https://github.com/AffiliateWP/AffiliateWP/pull/1836

DrewAPicture:
Let's punt #1836 to 2.0. There's a whole bunch of other places we could use the helper:

* `affiliate-wp.php`
* `affiliatewp/includes/class-affiliates-db.php`
* `affiliatewp/includes/class-creatives-db.php`
* `affiliatewp/includes/class-payouts-db.php`
* `affiliatewp/includes/class-referrals-db.php`
* `affiliatewp/includes/class-visits-db.php`
* `affiliatewp/includes/admin/AFFWP_Plugin_Updater.php`
* `affiliatewp/includes/admin/class-addon-updater.php`
* `affiliatewp/includes/admin/tools/tools.php`
* `affiliatewp/tests/test-rest.php`

PR #1836, opened 2016-11-23
issue/1834: Introduce compare_wp_version method

ramiabraham:
- Avoids direct call to `$wp_version` global
- Returns static boolean

#1834


DrewAPicture:
Removing from the 2.0 milestone. I think we should consider whether this new layer of abstraction has adequate utility. For instance, I think it would be a mistake to implement this with some kind of default operator as its usefulness is considerably depleted. Then if we add an operator parameter, the only remaining utility is that pulling the stored version is handled, but we're still supplying one, possibly two parameters for what is normally a 3 parameter function call.

Punting.

Issue #1468, opened 2016-07-22
Ability to add Notes against affiliates


wordmaid:
It would be great to be able to store notes against affiilates which are only visible to site admins. This could be a simple text area in the Edit Affiliate screen titled "Notes".

This could be used for: 
1. Promotion methods if the affiliate has been manually added
2. To store any specific payout information
3. Storing additional website URLs where the affiliate promotes the site owner's website, other than the primary URL
4. Storing links to social media accounts
5. Storing the affiliate's private contact details, such as phone number
6. Making notes regarding the affiliate for other site admins to be aware of
7. Making personal notes regarding the affiliate.

The possible uses are almost endless!

pippinsplugins:
This would be especially great for hands-on affiliate managers.

We could easily store notes in the meta table with one note per row.

sumobi:
@AffiliateWP/core Ready for testing (I have tested)

sumobi:
Oops, forgot to add it to the add affiliate screen

sumobi:
Reduced the height of the textarea by half.

<img width="1197" alt="screen shot 2016-12-10 at 5 38 17 pm" src="https://cloud.githubusercontent.com/assets/52581/21076682/d2c2574e-beff-11e6-9344-2a3171a1d55c.png">

pippinsplugins:
All works well here!

pippinsplugins:
The `test_add_affiliate_with_notes()` test is failing but I can't see why. Notes definitely get recorded during the affiliate creation (at least in the UI).

DrewAPicture:
@pippinsplugins I don't see any logic in `affwp_add_affiliate()` for actually setting the `notes` data in meta. Only in `affwp_update_affiliate()`. I think the test is failing because the data is actually never getting written.

pippinsplugins:
@DrewAPicture it's done by hooking into `affwp_insert_affiliate`. See `affwp_process_add_affiliate_meta()`

DrewAPicture:
In that case, the action is never firing because admin/affiliates/actions.php is never loaded during the test run (it's only loaded on the affiliates admin screens. So we'll have to bring in the file for that action to fire.

DrewAPicture:
1eb62c0 should fix it.

pippinsplugins:
Ah!

PR #1850, opened 2016-12-10
Issue/1468 - Add notes field

sumobi:
First pass #1468


includes/affiliate-functions.php
```diff
@@ -1089,10 +1090,20 @@ function affwp_update_affiliate( $data = array() ) {
 	do_action( 'affwp_updated_affiliate', affwp_get_affiliate( $affiliate ), $updated );
 
 	if ( $updated ) {
+
 		// Update affiliate's account email
 		if ( wp_update_user( array( 'ID' => $user_id, 'user_email' => $args['account_email'] ) ) ) {
```
sumobi: This is a bit weird to me, it seems like it's always having to update the affiliate's email address, regardless of whether it changed or not. For now, this is the only place I can see to update the affiliate meta, since if it doesn't return true, nothing will be updated.

includes/admin/affiliates/edit.php
```diff
@@ -190,6 +191,19 @@
...
 			</tr>
 
+			<tr class="form-row">
+
+				<th scope="row">
+					<label for="notes"><?php _e( 'Affiliate Notes', 'affiliate-wp' ); ?></label>
+				</th>
+
+				<td>
+					<textarea name="notes" rows="10" cols="50" id="notes" class="large-text"><?php echo esc_html( $notes ); ?></textarea>
```
sumobi: This textarea follows the same textareas that WP uses. I propose we change our "promotion method" textarea to follow suit since it's tiny and inconsistent.

includes/admin/affiliates/actions.php
```diff
@@ -34,6 +34,23 @@ function affwp_process_add_affiliate( $data ) {
 }
 add_action( 'affwp_add_affiliate', 'affwp_process_add_affiliate' );
 
+/**
+ * Add affiliate meta
+ *
+ * @since 2.0
+ * @return void
+ */
+function affwp_process_add_affiliate_meta( $affiliate_id ) {
```
pippinsplugins: Yeah I think this is okay.

pippinsplugins: Let's actually change this a bit and detect `notes` in the `$args` array, that way it's easier to do programmatically.

sumobi: done

sumobi:
@pippinsplugins updated ^

Issue #1443, opened 2016-07-18
Affiliate Form Settings


sumobi:
This issue has been a combined with 2 other issues which go hand in hand. The other issues have been closed for clarity. All discussion will be held here.

The old Issues that have been combined with this one:

https://github.com/AffiliateWP/AffiliateWP/issues/931 - Add Forms tab to settings
https://github.com/AffiliateWP/AffiliateWP/issues/1051 - Easily show or remove affiliate registration & affiliate login forms on affiliate area page

This issue covers:

- A filter to hide the registration and login form from the [affiliate_area] shortcode
- A new select menu in the admin settings to allow the user to control what forms are shown on the Affiliate Area page
- New checkbox options for selecting which affiliate registration fields should be required or not.

![screen shot 2016-12-27 at 4 25 21 pm](https://cloud.githubusercontent.com/assets/52581/21489573/18300f66-cc51-11e6-9eec-98b2d7f92702.png)

pippinsplugins:
:+1:

sumobi:
Oh check that out https://github.com/AffiliateWP/AffiliateWP/issues/1051. A filter to programmatically hide it and an admin dropdown would be helpful.

DrewAPicture:
I like #1051 as well. I think we should focus our attention there on the 1.9 milestone. @AffiliateWP/core  Any objections?

pippinsplugins:
No objections here.

pippinsplugins:
#1051 was moved to 2.0 so moving this as well.

sumobi:
@AffiliateWP/core Ready for review. I'd love to hear any thoughts on how to make it cleaner.

sumobi:
To test you can either play with the filters directly, setting them individually to `false` which will hide each form, or you can now change the select menu from the bottom of the Affiliates -> Settings -> General tab (location to be discussed).

sumobi:
### The "Affiliate Area Forms" select menu
Previous discussion was here: https://github.com/AffiliateWP/AffiliateWP/issues/1051
___

![screen shot 2016-12-27 at 3 57 01 pm](https://cloud.githubusercontent.com/assets/52581/21489263/3d1b28fa-cc4d-11e6-9662-bef35dc42480.png)

![screen shot 2016-12-27 at 3 57 08 pm](https://cloud.githubusercontent.com/assets/52581/21489262/3d1a9de0-cc4d-11e6-99ec-a75d5fdcba1c.png)

[...]

PR #1855, opened 2016-12-11
Issue/1443 - Add filters to programmatically show/hide the login and registration forms

sumobi:
#1443

Add filters to programmatically show/hide the login and registration forms


includes/class-shortcodes.php
```diff
@@ -31,30 +31,25 @@ public function affiliate_area( $atts, $content = null ) {
 
 		affwp_enqueue_script( 'affwp-frontend', 'affiliate_area' );
 
+		$show_registration = apply_filters( 'affwp_affiliate_area_show_registration', true ) ? affiliate_wp()->templates->get_template_part( 'register' ) : '';
```
pippinsplugins: `get_template_part` does not return a value, it outputs it, so this won't quite work. The `get_template_part` needs to be called down lower (like it always has been) but only if `$show_registration` is `true`.

sumobi: @pippinsplugins Hmm works for me (showed you this earlier via the gist) :) Happy to do it the other way though

includes/misc-functions.php
```diff
@@ -857,3 +857,37 @@ function affwp_enqueue_script( $handle, $context = '' ) {
...
 }
+
+
+/**
+ * Controls what forms are shown on the Affiliate Area page.
+ *
+ * @since  2.0
+ * @return void
+ */
+function affwp_affiliate_area_forms() {
```
sumobi: @pippinsplugins Is there anywhere else we should put this function?

includes/class-shortcodes.php
pippinsplugins: Let's go ahead and adjust it please.

includes/misc-functions.php
pippinsplugins: I don't think so but let's change the change. Perhaps `affwp_filter_shown_affiliate_area_forms()`?

sumobi: Agreed

sumobi:
Updated PR

includes/admin/class-upgrades.php
```diff
@@ -439,5 +443,24 @@ private function v195_upgrade() {
...
+	 *
+	 * @since 2.0
+	 * @access private
+	 */
+	private function v2_upgrade() {
+
+		// Update settings
+		affiliate_wp()->settings->set( array(
+			'required_registration_fields' => array(
+				'your_name'   => 'Your Name',
```
pippinsplugins: We don't need to set the values, just the keys. Pretty sure anyway.

sumobi: @pippinsplugins Just tried without values but the checkboxes wouldn't become checked by default. Looks like it needs the values.

pippinsplugins: :+1:

includes/admin/class-upgrades.php
```diff
@@ -452,7 +452,19 @@ private function v195_upgrade() {
 	private function v20_upgrade() {
 		@affiliate_wp()->capabilities->add_caps();
 		$this->log( 'Upgrade: Core capabilities have been upgraded.' );
+
+
+		// Update settings
+		affiliate_wp()->settings->set( array(
```
DrewAPicture: Let's silence any potential errors here with the `@` for consistency with the other upgrade routines.

includes/admin/settings/class-settings.php
```diff
@@ -36,6 +35,11 @@ public function __construct() {
 		// Capabilities
 		add_filter( 'option_page_capability_affwp_settings', array( $this, 'option_page_capability' ) );
 
+		// Filter the general settings
+		add_filter( 'affwp_settings_general', array( $this, 'required_registration_fields' ) );
```
DrewAPicture: Why are we filtering in _these_ specific settings while defining all others directly in class-settings.php?

includes/class-register.php
```diff
@@ -230,6 +231,54 @@ public function required_fields() {
...
 	}
 
+
+
+	/**
+	 * Make fields required/not required, based on the "Required Registration Fields" admin setting
+	 *
+	 * @since 2.0
+	 */
+	public function maybe_required_fields( $required_fields ) {
```
DrewAPicture: `$required_fields` and the return value should be documented here.

includes/class-shortcodes.php
```diff
@@ -31,29 +31,36 @@ public function affiliate_area( $atts, $content = null ) {
 
 		affwp_enqueue_script( 'affwp-frontend', 'affiliate_area' );
 
+		$show_registration = apply_filters( 'affwp_affiliate_area_show_registration', true );
```
DrewAPicture: Missing hook docs. If these are duplicates or permutations of a dynamic hook, still should point to the original hook that's actually documented, i.e. `/** This filter is documented in path/to/file.php */`

includes/class-shortcodes.php
```diff
@@ -31,29 +31,36 @@ public function affiliate_area( $atts, $content = null ) {
...
 		ob_start();
 
 		if ( is_user_logged_in() && affwp_is_affiliate() ) {
-
 			affiliate_wp()->templates->get_template_part( 'dashboard' );
-
 		} elseif ( is_user_logged_in() && affiliate_wp()->settings->get( 'allow_affiliate_registration' ) ) {
 
-			affiliate_wp()->templates->get_template_part( 'register' );
+			if ( $show_registration ) {
```
DrewAPicture: Let's make the boolean check for both of these filterable values explicit, i.e. `if ( true === $show_registration )`

includes/install.php
```diff
@@ -40,15 +40,20 @@ function affiliate_wp_install() {
 
 		// Update settings.
 		$affiliate_wp_install->settings->set( array(
-			'affiliates_page' => $affiliate_area
+			'affiliates_page'              => $affiliate_area,
+			'required_registration_fields' => array(
+				'your_name'   => 'Your Name',
```
DrewAPicture: Should these strings be translatable?

includes/admin/settings/class-settings.php
sumobi: Because the checkboxes should only show when Affiliate Registration is enabled from the Misc tab

includes/install.php
sumobi: Yep let's do that

includes/class-register.php
sumobi: Done

includes/class-shortcodes.php
sumobi: Done

includes/install.php
sumobi: Done

includes/admin/class-upgrades.php
sumobi: done

includes/class-shortcodes.php
sumobi: added doc blocks

DrewAPicture:
@sumobi Good to go here. Just waiting on tests.

PR #1864, opened 2016-12-14
issues/843: Encode space characters in affiliate URL generator

### LINKED ISSUE #843: Account for spaces in affiliate usernames in affiliate URLS
If an affiliate has a WordPress username with a space in it, eg `Dat Guy` then their affiliate referral URL won't work: 

`http://mysite.com/ref/Dat Guy`

or

`http://mysite.com/ref/Dat%20Guy`

WordPress itself doesn't let you register a username with a space in it however our affiliate registration form does. We should prevent spaces in usernames.

ref: https://secure.helpscout.net/conversation/121113989/17016/

screen shot 2015-09-17 at 7 50 14 pm

ramiabraham:
- Account for spaces in affiliate usernames in affiliate URLS

- Fixes #843

### DIFF
--- a/assets/js/frontend.js
+++ b/assets/js/frontend.js
@@ -1,144 +1,144 @@
-jQuery(document).ready(function ($) {
+jQuery(document).ready( function($) {
 
-  // Remove whitespace from the campaign name.
-  $('#affwp-campaign').on('focusout', function (event) {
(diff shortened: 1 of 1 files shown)


DrewAPicture:
@pippinsplugins I've restored the original indentations to tabs.

Issue #1629, opened 2016-09-22
PayPal: add support for rejecting referrals on refund


pippinsplugins:
The PayPal integration needs to support rejecting referrals when a refund IPN is received.

pippinsplugins:
Ready for testing!

pippinsplugins:
Had to make a few tweaks but it's fully working now. Will merge once tests pass.

PR #1868, opened 2016-12-15
Add auto rejection for PayPal integration. #1629

pippinsplugins:
#1629

### DIFF
--- a/includes/integrations/class-paypal.php
+++ b/includes/integrations/class-paypal.php
@@ -10,7 +10,7 @@ class Affiliate_WP_PayPal extends Affiliate_WP_Base {
 	 */
 	public function init() {
 
-		$this->context = 'paypal'; 
+		$this->context = 'paypal';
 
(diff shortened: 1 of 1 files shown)


includes/integrations/class-paypal.php
```diff
@@ -304,7 +304,7 @@ private function verify_ipn( $post_data ) {
 
 		$verified = false;
 		$endpoint = array_key_exists( 'test_ipn', $post_data ) ? 'https://www.sandbox.paypal.com/cgi-bin/webscr' : 'https://www.paypal.com/cgi-bin/webscr';
-		$args     = array_merge( array( 'cmd' => '_notify-validate' ),  $post_data );
+		$args     = array_map( 'stripslashes', array_merge( array( 'cmd' => '_notify-validate' ),  $post_data ) );
```
DrewAPicture: To simplify, we could also probably just use `wp_unslash()` here.

pippinsplugins: Updated

Issue #1280, opened 2016-05-11
Image thumbnail in Affiliates -> Creatives list table


DrewAPicture:
Creatives are inherently visual. And since they have the option of associating with an image, it might be nice to show a thumbnail (or fallback) in the list table to indicate which ones have images, and a thumbnail of that image if set.

sumobi:
@ramiabraham I can take this over if you'd like

DrewAPicture:
@sumobi Do eet.

sumobi:
@AffiliateWP/core Ready for testing

<img width="1463" alt="screen shot 2016-12-15 at 7 53 35 pm" src="https://cloud.githubusercontent.com/assets/52581/21251159/38c4a714-c300-11e6-82d0-7c77d6dd2e7e.png">

Rather than loading the entire image into the list table, I opted to load the thumbnail. Primarily because:

1. All images are the same size and it will be less visually distracting. Also, if a skyscraper banner is used, each row will still take up the same amount of space
2. It will be faster to load the thumbnail as opposed to the full image size

The only downside that I can see is that the thumbnail isn't a full representation of the creative since the creative could be a very specific size, e.g. 480x60.

It's only meant to be a preview though so I don't think this is too big of an issue Anyone agree/disagree? I also decided we don't need a placeholder image if there isn't an image added to a creative.

pippinsplugins:
How about changing the `Image` column name to `Image Preview`? I think that would adequately take care of concerns about it not showing the complete image.

sumobi:
@pippinsplugins sounds good to me

sumobi:
@pippinsplugins updated

pippinsplugins:
@sumobi images aren't showing up for me. It's just a blank space.

sumobi:
@pippinsplugins Do you have an image added to the creative? (of course you have)

Thoughts on a filter or option to hide the entire `image preview` column? I personally would prefer not to have this new column since it makes each row exponentially.

pippinsplugins:
Yep, definitely uploaded an image. I tried it with an existing creative and a brand new one with a new image upload.

:+1: to adding a new filter option.

sumobi:
Ok I'll have a look, I also tested with new creatives etc and it worked fine. Bet it's something small.

sumobi:
@pippinsplugins what's your thumbnail size set to in media -> settings? You do see the column though right? It's just empty. Do you see anything at all in the source for each `td`?

[...]

PR #1869, opened 2016-12-16
Issue/1280 - Add image preview to main list table

sumobi:
#1280


includes/admin/creatives/class-list-table.php
```diff
@@ -169,6 +170,25 @@ function column_url( $creative ) {
...
+	 *
+	 * @access public
+	 * @since 2.0
+	 * @return string image src of creative
+	 */
+	function column_image( $creative ) {
+		global $wpdb;
+
+		// Get the creative's attachment ID based on the image URL
+		$attachment = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE guid = '%s';", $creative->image ) );
```
pippinsplugins: This should use the attachment_url_to_postid() function: https://developer.wordpress.org/reference/functions/attachment_url_to_postid/

sumobi: Fixed

Issue #1787, opened 2016-11-03
Integration: Caldera Forms


michaelbeil:
It might be a good idea to add an integration for Caldera Forms. 

Ref: https://secure.helpscout.net/conversation/272708245/45996/?folderId=205524

sumobi:
Do we still want to do this for 2.0? If so, we'll need to get some of their payment add-ons (probably just Stripe): https://calderaforms.com/caldera-forms-add-ons/#/payment

pippinsplugins:
@sumobi could you request copies of them? If we can get it done, awesome, otherwise we can punt.

sumobi:
I've reached out to them

sumobi:
Added PR - WIP

sumobi:
Yeahhh tag team

pippinsplugins:
Pinging @Shelob9. The notes above may interest you :)

pippinsplugins:
@sumobi can you please put the add-ons in Dropbox?

Shelob9:
> Setting referrals to "unpaid" on a successful payment
> 
> There's no generic "payment completed" hook (like Gravity Forms)
> There are 4 payment gateways ("processors") for Caldera Forms. Of those, only Stripe and Braintree have payment completed hooks. PayPal and Authorize.net do not. For now we'll have 2 write 2 methods to mark the referral as unpaid.

This is correct. We do not have [this concept](https://github.com/CalderaWP/Caldera-Forms/issues/1076). The [caldera_forms_submit_complete](https://calderaforms.com/doc/caldera_forms_submit_complete/) hook will run when a submission has passed all validation and is about to be saved. 

For Paypal, in the Callback on that hook, do this:
```
    global $transdata;
    if( ! empty( $transdata['paypal_express'] ) && ! empty( $transdata['paypal_express']['result'] ) ){
        //$transdata['paypal_express']['result'] is info returned from PayPal about success
    
    }elseif( ! empty( $transdata['note'] ) ){
        //$transdata['note'] is error from paypal
    }else{
        //Things went really wrong
    }
```

For authorize.net there really is no good way to do this. @shelob9 hangs head in shame.

@sumobi What is your deadline on this? I can add hooks to each of them to make your life easy -- but that would mean your integration would require the new version of the add-ons. If you want to do that, please let me know what data you need in the callback.

> Linking the reference (entry ID) to the standalone entry page.
> 
[...]

PR #1877, opened 2016-12-28
Issue/1787 - Caldera Forms integration WIP

sumobi:
#1787


includes/integrations/class-caldera-forms.php
```diff
@@ -0,0 +1,119 @@
...
+
+		$affiliate_id = $this->affiliate_id;
+
+		// Return if the customer was not referred or the affiliate ID is empty
+		if ( ! $this->was_referred() && empty( $affiliate_id ) ) {
+			return;
+		}
+
+		// Prevent referral creation unless referrals enabled for the form
+		if ( ! $form['affwp_allow_referrals'] ) {
```
pippinsplugins: Let's use `! empty( $form['affwp_allow_referrals'] )` just in case the index isn't set.

sumobi: Updated

sumobi: Should actually be `empty( $form['affwp_allow_referrals'] )` or it will return if the form does allow referrals, updated.

includes/integrations/class-caldera-forms.php
```diff
@@ -0,0 +1,278 @@
...
+
+		$this->context = 'caldera-forms';
+
+		add_action( 'caldera_forms_submit_complete', array( $this, 'submit_complete' ), 10, 3 );
+		add_action( 'caldera_forms_general_settings_panel', array( $this, 'add_settings' ) );
+
+		add_action( 'caldera_forms_pre_load_processors', array( $this, 'pre_load_processors' ), 10, 6 );
+		add_action( 'caldera_forms_includes_complete', array( $this, 'load_processor' ), 10, 6 );
+
+		add_action( 'cf_stripe_post_successful_charge', array( $this, 'complete_payment_stripe' ), 10, 4 );
```
Shelob9: @sumobi These actions are not needed if you use the processor to register conversions. Feel very redundant.

includes/integrations/class-caldera-forms.php
```diff
@@ -0,0 +1,162 @@
...
+	/**
+	 * Register the form-specific settings
+	 *
+	 * @since  2.0
+	 * @return void
+	 */
+	public function add_settings( $element ) {
+		?>
+
+		<div class="caldera-config-group">
```
Shelob9: @sumobi Please don't reverse engineer our processor UI's HTML. Integrations should use the `Caldera_Forms_Processor_UI` class. I intend to make changes to processor UI JavaScript in the middle of this year and those that use our old system may experience breaking changes. Those that use the proper API will not.

sumobi: @Shelob9 Can you provide an example of how to add a new checkbox to the form settings? I must be missing something, that class looks like it only interacts with the processor UI.

Issue #1878, opened 2016-12-29
Gravity Forms: Block referral if affiliate's email is used


sumobi:
Currently the Gravity Forms integration does not check the submitted email address(es) to see if it could be the affiliate's. We should block the referral much like our other integrations if this is the case.

sumobi:
@AffiliateWP/core Ready for testing (I have tested)

pippinsplugins:
@sumobi looks good. One small change on the PR please.

sumobi:
@pippinsplugins updated

PR #1879, opened 2016-12-29
Issues/1878 - Block referral if affiliate’s email is used

sumobi:
This checks all email fields on the Gravity Form, and if any email matches the affiliate’s email, the referral is blocked.

#1878

### DIFF
--- a/includes/integrations/class-gravityforms.php
+++ b/includes/integrations/class-gravityforms.php
@@ -268,11 +268,7 @@ class Affiliate_WP_Gravity_Forms extends Affiliate_WP_Base {
 			}
 		}
 
-		if ( $emails ) {
-			return $emails;
-		}
-
-		return false;
+		return $emails;
 
 	}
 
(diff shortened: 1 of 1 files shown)


includes/integrations/class-gravityforms.php
```diff
@@ -225,6 +248,35 @@ public function maybe_check_coupons( $form, $entry ) {
...
+		if ( $email_fields ) {
+			foreach ( $email_fields as $email_field ) {
+				if ( ! empty( $entry[ $email_field->id ] ) ) {
+					$emails[] = $entry[ $email_field->id ];
+				}
+			}
+		}
+
+		if ( $emails ) {
+			return $emails;
```
pippinsplugins: Let's have it always return `$emails` and never `false`. If no emails are found, it will be an empty array.

sumobi: Good catch, updated

Issue #1885, opened 2017-01-03
Add setting to opt into beta releases


pippinsplugins:
With EDD Software Licensing 3.5 now running on the site, it's possible for us to release beta versions.

We should add a new setting so customers can opt into beta versions.

PR #1887, opened 2017-01-04
Issue/1885 – Add setting to opt in to beta releases

pippinsplugins:
#1885

### DIFF
--- a/affiliate-wp.php
+++ b/affiliate-wp.php
@@ -482,7 +482,7 @@ final class Affiliate_WP {
 				'item_name' => 'AffiliateWP',
 				'item_id'   => 17,
 				'author'    => 'Pippin Williamson',
-				'beta'      => $this->settings->get( 'betas' )
+				'beta'      => $this->settings->get( 'betas', false )
 			)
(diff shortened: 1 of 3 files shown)


affiliate-wp.php
```diff
@@ -477,11 +477,12 @@ private function updater() {
...
 		$affwp_updater = new AFFWP_Plugin_Updater( 'https://affiliatewp.com', __FILE__, array(
-				'version' 	=> AFFILIATEWP_VERSION,
-				'license' 	=> $license_key,
+				'version'   => AFFILIATEWP_VERSION,
+				'license'   => $license_key,
 				'item_name' => 'AffiliateWP',
 				'item_id'   => 17,
-				'author' 	=> 'Pippin Williamson'
+				'author'    => 'Pippin Williamson',
+				'beta'      => $this->settings->get( 'betas' )
```
DrewAPicture: Think maybe we should supply a default here, just in case?

pippinsplugins: Added.

PR #1888, opened 2017-01-04
Issue/1500 - Stripe integration through WP Simple Pay

pippinsplugins:
#1500


includes/class-integrations.php
```diff
@@ -26,6 +26,7 @@ public function get_integrations() {
 			'pmp'            => 'Paid Memberships Pro',
 			'pms'            => 'Paid Member Subscriptions',
 			'rcp'            => 'Restrict Content Pro',
+			'stripe'         => 'Stripe (through WP Simple Pay)',
```
sumobi: @pippinsplugins This needs to be moved down the list a bit

includes/integrations/class-stripe.php
```diff
@@ -0,0 +1,123 @@
+<?php
+
+class Affiliate_WP_Stripe extends Affiliate_WP_Base {
+
+	/**
+	 * Get thigns started
```
sumobi: Typo

Issue #1046, opened 2016-01-08
Generate a payout for a single affiliate only


flowdee:
PR: #1889

Right now it's possible to filter referrals for a single user. But it's not possible to generate the total amount of referrals and mark them as paid (for a specific date range) for a single user as well. 

With the tools I can export the referrals of a single user but there is no total amount.

pippinsplugins:
This would be nice.

pippinsplugins:
@flowdee While the proposed changes in https://github.com/AffiliateWP/AffiliateWP/pull/1047 are good, we really need to support setting one or more affiliates instead of only permitting a single affiliate.

flowdee:
Indeed @pippinsplugins. This should affect the tools (Export Referrals) too I guess.

pippinsplugins:
It would be nice to add it to both.

DrewAPicture:
Added a PR: #1889

![Affiliate field](https://cl.ly/2a0b0611333E/Screen%20Shot%202017-01-04%20at%205.25.00%20PM.png)

pippinsplugins:
@DrewAPicture This needs to use the batch processor as well from #1707.

pippinsplugins:
To clarify, #1707 already added the batch processor to the payout file generation, we just need to apply the affiliate-specific payouts to the new batch processing instead of the old export method.

DrewAPicture:
Yup, on it :-)

DrewAPicture:
@pippinsplugins Actually, that's already been done in the batch process. I didn't even convert the legacy exporter to support a single affiliate.

DrewAPicture:
@pippinsplugins Single affiliate support has been added to `Generate_Payouts` in the PR: #1889

pippinsplugins:
Excellent!

flowdee:
Awesome 👍

pippinsplugins:
This works well except for a couple of issues that I have noted on #1707

DrewAPicture:
@pippinsplugins I've addressed the issues you raised on #1707.

pippinsplugins:
Good here!

pippinsplugins:
Ready for merge!

DrewAPicture:
Updated following merge of #1156 to use `user_name` only and not rely on `user_id`. Retested and looks good. Waiting for tests to pass.

DrewAPicture:
Merged #1889.

PR #1889, opened 2017-01-05
issue/1046 – Add the ability to generate a payout for a single affiliate only.

DrewAPicture:
Issue: #1046


pippinsplugins:
@DrewAPicture looks like we have a merge conflict now.

Issue #338, opened 2014-08-15
CSV import for affiliates and referrals


pippinsplugins:
We should provide a way to import affiliates and referrals via a CSV file.

Related #337 
PR: #1891

pippinsplugins:
This should be supported in core and is something we definitely want to support so re-opening.

sumobi:
Another request for CSV importing. Customer is coming over from https://www.getambassador.com/.

Having a CSV importer for affiliates/referrals would be so helpful. Similar to EDD's, I'd imagine we just map the column data to specific values inside AffiliateWP.

pippinsplugins:
Let's try and do it for 2.0.

wordmaid:
Along the same lines, this customer wants to be able to import affiliates: https://secure.helpscout.net/conversation/256116186/43271/?folderId=205524

DrewAPicture:
Definitely. It would be nice if we could (at the very least) make our exports importable.

wordmaid:
Yes! 👍

swise23:
We also need the import affiliates option.

DrewAPicture:
I started a PR that we can launch off of when we're ready in 2.1: #1891

TODO:

* [ ] Build CSV field mapping UI a la EDD
* [ ] Implement referral imports via extending `Batch\Import` and related CSV interfaces
* [ ] Implement affiliate imports via extending `Batch]Import` and related CSV interfaces

DrewAPicture:
@AffiliateWP/core  This is now ready for testing.

Note that this PR implements the _architecture_ for batch imports as well as the _affiliate importer_ rather than the _referral importer_. Once we've worked out some of the kinks, it should be fairly simple work to do referral imports.

To test, you're looking for the Import Affiliates box in Tools > Import/Export:
![Import Affiliates](https://cl.ly/2X0Y2O2D3x3F/Screen%20Shot%202017-04-01%20at%2012.00.48%20AM.png)

Notable feedback:
* Any JavaScript errors
* Any fields not properly mapped
* Any fields you feel like are missing (remember, we're using our own affiliate exports as a model)
* Any breakage with any already-existing batch processes

Expectations:
* If the user already exists and is an affiliate, it will be skipped
* New users will be created using the Username value first, followed by a sanitized version of the email, followed by a generic affiliate####
* Statuses for new affiliates will follow existing logic in `affwp_add_affiliate()`, which means that if approval is required, they'll be set to "pending" etc.
[...]

PR #1891, opened 2017-01-05
Issue/338 – Batch importer architecture (WIP)

DrewAPicture:
Issue: #338

This stubs out a PR for introducing batch import architecture. It is built to extend functionality introduced in v2.0 and will be tackled at a later date.


includes/admin/ajax-actions.php
```diff
@@ -139,7 +139,18 @@ function affwp_process_batch_request() {
 	 *
 	 * @var \AffWP\Utils\Batch_Process\Export|\AffWP\Utils\Batch_Process\Base $process
 	 */
-	$process = new $class( $step );
+	if ( isset( $_REQUEST['data']['upload']['file'] ) ) {
+
+		// If this is an import, instantiate with the file and step.
+		$file = sanitize_text_field( $_REQUEST['data']['upload']['file'] );
+		$process = new $class( $file, $step );
```
pippinsplugins: Could we have a class_exists sanity check here?

includes/admin/tools/import/class-batch-import-affiliates.php
```diff
@@ -89,4 +192,64 @@ public function get_message( $code ) {
...
+				$first_name = sanitize_text_field( $args['first_name'] );
+			}
+
+			if ( ! empty( $args['last_name'] ) ) {
+				$last_name = sanitize_text_field( $args['last_name'] );
+			}
+
+			$user_id = wp_insert_user( array(
+				'user_login' => sanitize_user( $args['payment_email'], $this->use_strict ),
+				'user_email' => sanitize_email( $args['payment_email'] ),
```
pippinsplugins: Let's not use sanitize_email() here as it strips out some characters, like `+` that are valid characters. Use `sanitize_text_field()` instead. See https://core.trac.wordpress.org/ticket/21537

includes/admin/ajax-actions.php
DrewAPicture: There's a `class_exists()` check on `$class` a few lines up on L#126.

includes/admin/tools/import/class-batch-import-affiliates.php
DrewAPicture: Roger.

Issue #1907, opened 2017-01-11
Filter referrals on a per-affiliate basis for specific date range


wordmaid:
PR: #1924

Currently it’s not possible to filter referrals on a per-affiliate basis for a specific date range. Thoughts on allowing this somehow, such as adding an auto-complete search box to search for an affiliate's name next to the date range filter?

DrewAPicture:
Added a PR: #1907

pippinsplugins:
PR works great. Ready to merge once tests pass.

PR #1924, opened 2017-01-15
issue/1907 – Add an affiliate filter field to the Referrals list table.

DrewAPicture:
Issue: #1907


pippinsplugins:
Works great.

PR #1950, opened 2017-01-24
Release/2.0

pippinsplugins:


includes/admin/class-upgrades.php
```diff
@@ -34,69 +62,88 @@ public function __construct() {
...
 			$this->v19_upgrade();
 		}
 
-		if ( version_compare( $version, '1.9.5', '<' ) ) {
+		if ( version_compare( $this->version, '1.9.5', '<' ) ) {
 			$this->v195_upgrade();
 		}
 
+		// TODO: switch && to || before final release.
+		if ( version_compare( $this->version, '2.0', '<' ) && ! affwp_has_upgrade_completed( 'upgrade_v20_recount_unpaid_earnings' ) ) {
```
DrewAPicture: @pippinsplugins What do you think? Should we switch back to the `||` comparison operator here?

pippinsplugins: @DrewAPicture Yes, this should be an `||` I think.

PR #1954, opened 2017-01-25
Full v2.0 spanish translation

fernandotellado:
All the new strings for v2.0 and reviewed some other older strings


DrewAPicture:
@fernandotellado How this is different from your changes merged in #1959 and included in 2.0.1?

Issue #1967, opened 2017-01-30
Add "context" column in visits table


sumobi:
It would be helpful to have a new column in the visits DB table called `context`.

Original conversation:
https://github.com/AffiliateWP/affiliatewp-direct-link-tracking/issues/38#issuecomment-275898471

DrewAPicture:
Added a PR: #1969, which includes support for:

* [x] A new `context` field in the `AffWP\Visit` object
* [x] New `context` argument in `add()` and `update()`
* [x] New `context` and `context_compare` arguments in `get_visits()` + tests
* [x] A 2.0.2 upgrade routine to add the column

sumobi:
yeehaw, will test today 👍

DrewAPicture:
Still needed:

* [ ] Add `context` support to queries performed for the Visits list table
* [ ] WP-CLI support
* [ ] Tests for `add()` in regard to the `context` argument
* [ ] Tests for `update()` in regard to the `context` argument

sumobi:
Works great  👍

DrewAPicture:
I'll follow up here before the end of the week on the [remaining items](https://github.com/AffiliateWP/AffiliateWP/issues/1967#issuecomment-276211210).

DrewAPicture:
As of c757e4e7febed8e5c62594f99a9e22e25e32bf34, list table and wp-cli support has been added.

DrewAPicture:
And 4d3941f4497ebfbdc7e8ca9fa7bef069ee1b3e53 adds tests for `add()` and `update()` for `context`.

@AffiliateWP/core This is ready for final testing.

pippinsplugins:
@DrewAPicture noticed a couple of small items. See the PR review please.

DrewAPicture:
@pippinsplugins I've addressed your two points of feedback.

pippinsplugins:
Should we automatically set `context` to the value of `$_REQUEST['context']` if it is present?

PR #1969, opened 2017-01-30
Issue/1967 – Visit context column + tests

DrewAPicture:
Issue: #1967

Includes tests.


includes/REST/v1/class-visits-endpoints.php
```diff
@@ -201,6 +202,17 @@ public function get_collection_params() {
 			},
 		);
 
+		// 'visit_context' so as not to conflict with the global 'content' parameter.
+		$params['visit_context'] = array(
+			'description'       => __( 'The context under which the visit was created.', 'affiliate-wp' ),
+			'sanitize_callback' => function( $param, $request, $key ) {
+				return sanitize_key( substr( $param, 0, 50 ) );
+			},
+			'validatE_callback' => function( $param, $request, $key ) {
```
pippinsplugins: Incorrectly capitalized `E` here.

includes/admin/visits/class-list-table.php
```diff
@@ -287,7 +320,8 @@ public function visits_data() {
 		$user_id      = isset( $_GET['user_id'] )   ? absint( $_GET['user_id'] )               : false;
 		$referral_id  = isset( $_GET['referral'] )  ? absint( $_GET['referral'] )              : false;
 		$affiliate_id = isset( $_GET['affiliate'] ) ? absint( $_GET['affiliate'] )             : false;
-		$campaign     = isset( $_GET['campaign'] )  ? sanitize_text_field( $_GET['campaign'] ) : false;
```
pippinsplugins: I don't think `sanitize_key` is not correct for the campaign since affiliates could easily use spaces and other non-key-safe characters in their campaign names.

includes/REST/v1/class-visits-endpoints.php
DrewAPicture: Fixed in 663507f50ff4cd6b73be4376b05fce2d1cac4204.

includes/admin/visits/class-list-table.php
DrewAPicture: Change reverted in 1ef377c3acd38758de5d83ff247d2b1706b08efc.

Issue #1981, opened 2017-02-02
Add Affiliate_WP_Integrations method to check if a single integration is enabled


ramiabraham:

ramiabraham:
Added a PR (defines a few other minor related items): https://github.com/AffiliateWP/AffiliateWP/pull/1982

PR #1982, opened 2017-02-02
Issue/1981: Add methods to check single integrations

ramiabraham:
- Defines `Affiliate_WP_Integrations::integration_is_enabled()`
- Defines `Affiliate_WP_Integrations::integration_is_valid()`
- Related inline docs for `Affiliate_WP_Integrations` properties.
- Implements #1981

### DIFF
--- a/includes/class-integrations.php
+++ b/includes/class-integrations.php
@@ -8,8 +8,22 @@ class Affiliate_WP_Integrations {
 
 	}
 
+	/**
+	 * Gets all available AffiliateWP integrations
+	 *
+	 * @since  1.0
+	 *
+	 * @return array $integrations  Available AffiliateWP integrations.
+	 */
 	public function get_integrations() {
 
+		/**
+		 * An array of all available AffiliateWP integrations.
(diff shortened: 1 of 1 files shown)


includes/class-integrations.php
```diff
@@ -43,10 +57,56 @@ public function get_integrations() {
...
+	 *
+	 * @param  string  $integration  The integration to check.
+	 * @return boolean               True if a valid integration, otherwise false.
+	 * @since  2.1
+	 */
+	public function integration_is_valid( $integration ) {
+		$all_integrations = $this->get_integrations();
+
+		if ( array_key_exists( $integration, $all_integrations ) ) {
+			return true;
```
pippinsplugins: Let's combine these return statements into a single `return array_key_exists( $integration, $all_integrations )` and add a filter. Same thing below for the enabled method.

ramiabraham: Updated in https://github.com/AffiliateWP/AffiliateWP/pull/1982/commits/280643ffb738cc3426df884f0af38c7d617140a7

PR #1990, opened 2017-02-08
Replaced generate_random_password deprecated function

tubiz:
I was working on migrating affiliates data from WP Affiliate to AffiliateWP and noticed that the migration was taking too long.

After checking the log I saw the error below

```
[08-Feb-2017 09:57:21 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function AffWP\Utils\Batch_Process\generate_random_password() in /Applications/MAMP/htdocs/nbp/wp-content/plugins/affiliate-wp/includes/admin/utilities/class-batch-migrate-wp-affiliate.php:155
Stack trace:
#0 /Applications/MAMP/htdocs/nbp/wp-content/plugins/affiliate-wp/includes/admin/ajax-actions.php(156): AffWP\Utils\Batch_Process\Migrate_WP_Affiliate->process_step()
#1 /Applications/MAMP/htdocs/nbp/wp-includes/class-wp-hook.php(298): affwp_process_batch_request('')
#2 /Applications/MAMP/htdocs/nbp/wp-includes/class-wp-hook.php(323): WP_Hook->apply_filters('', Array)
#3 /Applications/MAMP/htdocs/nbp/wp-includes/plugin.php(453): WP_Hook->do_action(Array)
#4 /Applications/MAMP/htdocs/nbp/wp-admin/admin-ajax.php(91): do_action('wp_ajax_process...')
#5 {main}
  thrown in /Applications/MAMP/htdocs/nbp/wp-content/plugins/affiliate-wp/includes/admin/utilities/class-batch-migrate-wp-affiliate.php on line 155
```

I have replaced the deprecated generate_random_password function with wp_generate_password


DrewAPicture:
@tubiz Nice catch! The interesting part is that the entire reason the fatal was thrown is that the global function is called within the `AffWP\Utils\Batch_Process` namespace. Meaning that yes, we should probably not use a deprecated function, but also that the function needs to be called with a global namespace, e.g. `\wp_generate_password()` :)

includes/admin/utilities/class-batch-migrate-wp-affiliate.php
```diff
@@ -152,7 +152,7 @@ public function process_step() {
 						'first_name' => $affiliate->firstname,
 						'last_name'  => $affiliate->lastname,
 						'user_url'   => $affiliate->website,
-						'user_pass'  => generate_random_password( 20 ),
+						'user_pass'  => wp_generate_password( 20 ),
```
DrewAPicture: Should be called in the global namespace, `\wp_generate_password( 20 )`.

DrewAPicture:
Or even more likely, ms-deprecated.php isn't loaded under the circumstances you saw the fatal. Either way, should probably use the global namespace.

Issue #2016, opened 2017-02-13
Add action hook to process_bulk_action() method in affiliates list table class


sumobi:
We should add an action hook in the `process_bulk_action()` method of the affiliates list table class like we do for the referrals list table class: https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/admin/referrals/class-list-table.php#L678

sumobi:
Added PR: https://github.com/AffiliateWP/AffiliateWP/pull/2017

pippinsplugins:
I don't see any issues with this :+1:

PR #2017, opened 2017-02-13
Issue/2016 - Add action hook to process_bulk_action()

sumobi:
and formatting #2016

### DIFF
--- a/includes/admin/affiliates/class-list-table.php
+++ b/includes/admin/affiliates/class-list-table.php
@@ -596,11 +596,11 @@ class AffWP_Affiliates_Table extends List_Table {
 	 */
 	public function process_bulk_action() {
 
-		if( empty( $_REQUEST['_wpnonce'] ) ) {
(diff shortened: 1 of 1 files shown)


includes/admin/affiliates/class-list-table.php
```diff
@@ -634,6 +634,16 @@ public function process_bulk_action() {
 				affwp_set_affiliate_status( $id, 'inactive' );
 			}
 
+			/**
+			 * Fires after an affiliate bulk action is performed.
+			 *
+			 * The dynamic portion of the hook name, `$this->current_action()` refers
+			 * to the current bulk action being performed.
+			 *
```
DrewAPicture: Needs an `@since` tag and version. A lot of the hook docs that are missing versions are ones we just haven't tracked down yet. ;)

sumobi:
Will add, I'm guilty as charged. I'm trying to be a better person.

Issue #2048, opened 2017-02-23
Remove checking the currently logged-in user in affwp_get_affiliate_id() in the admin


DrewAPicture:

DrewAPicture:
Added a PR with tests: #2050

PR #2050, opened 2017-02-23
issue/2048 – No logged-in-user in admin or doing Ajax

DrewAPicture:
Issue: #2048


TimothyBJacobs:
To avoid the constants issue, might it make sense to shim `wp_doing_ajax()` that way the filter can be used?

PR #2084, opened 2017-03-08
Update class-memberpress.php

harrisonsmith:
When the subscription ID was supplied as a non-array, wpdb->insert would fail as this would cause the insert_pending_referral() method ( includes\integrations\class-base.php:133 ) to insert a row with the column 'custom' as NULL (because maybe_serialize() would result in null being returned when just passing the subscription_id), and the column 'custom' is NOT NULL, so wpdb->insert throws an error:

Here's the error I captured in my error log:
[08-Mar-2017 16:06:18 UTC] WordPress database error Column 'custom' cannot be null for query INSERT INTO `wp_affiliate_wp_referrals` (`affiliate_id`, `date`, `currency`, `status`, `amount`, `reference`, `description`, `campaign`, `visit_id`, `products`, `custom`, `context`) VALUES (17, '2017-03-08 16:06:18', 'GBP', 'pending', '100.00', '59', 'Full Ukandoo Membership', '', 47, '', NULL, 'memberpress') made by require('wp-blog-header.php'), require_once('wp-load.php'), require_once('wp-config.php'), require_once('wp-settings.php'), do_action('init'), WP_Hook->do_action, WP_Hook->apply_filters, call_user_func_array, MeprAppCtrl::parse_standalone_request, MeprCheckoutCtrl->process_signup_form, MeprTransaction->store, MeprHooks::do_action, call_user_func_array, do_action('mepr-txn-status-pending'), WP_Hook->do_action, WP_Hook->apply_filters, call_user_func_array, Affiliate_WP_MemberPress->add_pending_referral, Affiliate_WP_Base->insert_pending_referral, Affiliate_WP_Referrals_DB->add, Affiliate_WP_DB->insert

Love the plugin guys!


DrewAPicture:
I'm not sure we're getting at the root of the problem with this proposed fix.

A null value wrapped in an array is still a null value, but it's now wrapped in an array, which isn't very useful down the line. That said, even if the value going in is literally `null`, it would still be considered empty and store an empty string rather than be passed through `maybe_serialize()`.

So I'm not sure about the situation where `maybe_serialize()` would be passing back a `null` value unless we were passing a null value in to begin with – and again, even if we were, it wouldn't pass the `! empty()` check which triggers `maybe_serialize()` in the first place. Which then of course begs the question of why it's null and if that's expected. 

What value _exactly_ do we expect to be passing in here?

harrisonsmith:
Hi! Thanks for the reply.

I found this error when attempting to integrate Lifetime Commissions with MemberPress, which I thankfully now have working.

The value that is inserted into the custom column after wrapping the value within an array was: 
a:2:{s:17:"lifetime_referral";b:1;i:0;i:29;}

The only addition I made to the Lifetime Commissions plugin was to add in the Memberpress class, in a new file named class-memberpress.php (within integrations folder). Which you can image was very simple as it's only one method that needs to return the ID or Email when given the reference (transaction ID). This may be the underlying issue perhaps?

I am not familiar enough with the AffiliateWP plugins to know the exact cause, but I hope the info above helps you get to the bottom of it! And if you do, please let me know as I'd be interested to know also.

Cheers!

DrewAPicture:
Can you link a gist/repo copy of class-memberpress.php file that you dropped into the lifetime commissions integrations directory?

harrisonsmith:
Sure thing :) - https://github.com/harrisonsmith/AffWP-Lifetime-Commissions

To be placed in affiliate-wp-lifetime-commissions\integrations\

Issue #2085, opened 2017-03-08
Restrict Content Pro: purchases creating referrals that shouldn't 


gspice:
This ticket has the history and admin account to log in:

https://secure.helpscout.net/conversation/326388477/54187/?folderId=1170842

Each membership subscription is creating a referral even though a discount code was not used.  They are not using affiliate links yet, only discount codes for their existing affiliates.  Currently any new purchase creates a referral for the affiliate id 27 (kc).  Previously it was another affiliate and they deleted it and it's now attached to this one.   

Also, the reference numbers in the Referral => Reference column do not link to anything.  They appear to attempt to link to Payments but the payment number is off by one.  Also searching the payments by any 3 digit payment id returns empty.  (not sure if that is a separate bug)

Referral created by user's test purchase this morning:

Member

https://bloomforwomen.com/wp-admin/admin.php?page=rcp-members&edit_member=6252

Invoice

https://bloomforwomen.com/wp-admin/index.php?payment_id=923&rcp-action=download_invoice&_wpnonce=77ab51e20f

Referral:

https://bloomforwomen.com/wp-admin/admin.php?page=affiliate-wp-referrals&referral_id=212&action=edit_referral

DrewAPicture:
Hmm. I logged in to the customer's site (in one browser) and double checked all settings and then switched over to sandbox mode so test Stripe CC would work.

Then in a second browser I registered two test members and neither generated referrals. So that's expected.

I can definitely see the referrals pointing to phantom payments in their admin though. I also noticed, for instance, that their settings are set to flat rate and $18 for recurring referrals, which seem to be the amount of the most recent referrals. Don't know if there's a connection there or not, but worth mentioning the correlation in case it's a clue.

Either way, @gspice what steps did you take _exactly_ to reproduce the problem?

gspice:
@DrewAPicture I did not do the test purchase from this morning , the customer did.  He purchased from the website page.  I was not sure how to go into test mode on Stripe to accomplish but I can revisit to see what I can find out.  Yes, you are correct, the rate for recurring referrals is $18 so it would seem these new purchases it believes are recurring payments?   

Regarding the reference payment id, in fairness, if you go to RCP's payment screen and type in a valid invoice number to search by it returns empty.  Not sure if that was searchable by the invoice number at one time, and is that a separate bug?

pippinsplugins:
@nosegraze could you lend a hand here?

gspice:
[...]

PR #2095, opened 2017-03-13
issue/2085 - return false in get_by() methods if the column or row_id values are eempty

pippinsplugins:
#2085


DrewAPicture:
Looks good. I added two tests each for the base and referrals db classes.

PR #2119, opened 2017-03-22
Issue/2120 Allow plugins to extend the login form with password alternatives

DavidAnderson684:
The previous code provides no easy/elegant way for an extension to extend the login form. In my use case, I am developing a plugin which will allow the user to scan a visible code with their smartphone, as a non-compulsory alternative to a password.

This patch provides a filter which allows an extension to instruction Affiliate-WP that it does not need to verify the password (which would be used in the case that the extension has performed its own verification).

Issue: #2120

### DIFF
--- a/includes/class-login.php
+++ b/includes/class-login.php
@@ -69,16 +69,20 @@ class Affiliate_WP_Login {
 			$this->add_error( 'no_such_user', __( 'No such user', 'affiliate-wp' ) );
 		}
 
-		if ( empty( $_POST['affwp_user_pass'] ) ) {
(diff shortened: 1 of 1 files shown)


DrewAPicture:
Interesting suggestion. Would you mind also adding documentation for the new hook?

michaelbeil:
@DavidAnderson684 this could be excellent!

DavidAnderson684:
@DrewAPicture I have done this now.

DrewAPicture:
Thanks!

Issue #2143, opened 2017-04-02
Remove background on code blocks inside creative HTML


sumobi:
With the TwentySixteen theme (and others), it's possible the creative code could look like this: 

![screen shot 2017-04-03 at 12 36 43 am](https://cloud.githubusercontent.com/assets/52581/24587172/bd957676-1805-11e7-9bcc-01ee66e56827.png)

This is simply because the `code` selector has a background color applied to it. As a customer pointed out, it looks like the HTML is already selected which can be confusing for the affiliate. It also doesn't look particularly amazing. 

I propose we set the background to none for code blocks inside creatives (and in case any other background styling has been applied). The code block can be colored if need be by applying a background color to the `pre` tag as we've already done.

End result:

![screen shot 2017-04-03 at 12 46 09 am](https://cloud.githubusercontent.com/assets/52581/24587241/f5969068-1806-11e7-81e8-a7c84e55b071.png)

sumobi:
PR: https://github.com/AffiliateWP/AffiliateWP/pull/2144

pippinsplugins:
@sumobi do you want to do borders and padding as well, as mentioned y @spencerfinnell?

sumobi:
@pippinsplugins @spencerfinnell

I think I prefer the keep the border and padding at least for now and just remove the background color on the `code` tag as per the PR. The background and border I feel would help affiliates see exactly what to copy and paste, especially when there's a long list of creatives. Obviously this depends on how the current theme has styled it.

Another small concern is a customer contacting us about how it looks without the border and padding. If they are a bit knowledgeable about CSS they might ask why the `pre` tag from their theme doesn't match the `pre` tag with the creatives. If they like how their theme uses `pre` tags they will have to add more CSS just to add the styling back. At least if the `pre` tag styling from their theme looks sub-par, we can say it's coming from their theme.

@spencerfinnell Would love to hear any more thoughts about removing the border and padding.

@AffiliateWP/core What does everyone prefer? 

Here's Twenty Sixteen with no border and padding (+ the change applied in PR)

![screen shot 2017-04-04 at 10 03 05 pm](https://cloud.githubusercontent.com/assets/52581/24651595/8e6429aa-1982-11e7-8bf6-dfd9b5d0192f.png)

And here's the default theme styling (+ the change applied in PR)

[...]

PR #2144, opened 2017-04-02
Issue/2143 - Remove background on code blocks inside creatives

sumobi:
Issue: #2143


spencerfinnell:
You may want to do borders and padding as well.

Issue #2160, opened 2017-04-13
Integrations: debug logging not writing messages


pippinsplugins:
The `log()` method of the integrations class isn't successfully writing the messages to the log file.

This happened some time after the global log utility was added.

PR #2161, opened 2017-04-13
Issue/2160

pippinsplugins:
Set debug property to correct value for backwards compatibility.
Remove debug check in log method.
Use utilities class in log() method.

Issue: #2160


DrewAPicture:
I don't see a need to retain checking `$this->debug` all over the place if `$this->log()` (via the global method) now checks it also. Absolutley retain the value for back-compat but it's now moot.

DrewAPicture:
Let's remove the `$this->debug` checks in 2.1.

pippinsplugins:
@DrewAPicture I agree. They should be removed from all integrations. I started to do it here but then decided to keep this PR limited to the bug fix itself.

DrewAPicture:
Added two tests to demonstrate `debug` property back-compat.

Issue #2101, opened 2017-03-15
Upgrade database notice appears on fresh installs


amdrew:
Installed AffiliateWP on a fresh sub-site of a MS installation and the upgrade database notice appears.

DrewAPicture:
Looks like some cruft from the merge into 2.0 was left over. `$this->version` was never actually assigned a value. Turns out 2.0 can't be less than `null`.

PR: #2102

sumobi:
This is still an issue. Just created a new sub-site and activated AffiliateWP 2.0.6:

<img width="1161" alt="screen shot 2017-04-20 at 1 17 35 pm" src="https://cloud.githubusercontent.com/assets/52581/25208910/cbacfc06-25cb-11e7-8e59-29ecd590a343.png">

DrewAPicture:
@sumobi Can you try installing a copy off of the `release/2.1` branch and confirm this is still happening?

sumobi:
@DrewAPicture Confirmed, still an issue with 2.1

DrewAPicture:
Does running the upgrade process clear the notice? This may just be a case of not setting the upgrade routine complete in newer versions.

sumobi:
After clicking upgrade database, you see:

![screen shot 2017-04-22 at 10 41 26 am](https://cloud.githubusercontent.com/assets/52581/25298379/5222b12c-2748-11e7-977b-ed71198e8181.png)

Clicking to dismiss the notice correctly removes it.

pippinsplugins:
@DrewAPicture if we could also include a tile that shows the average / total number of referrals in payouts, that would be swell.

DrewAPicture:
Simple PR: #2242 

@sumobi Could you please retest here?

pippinsplugins:
@DrewAPicture Logic looks good but there is a fatal error during install:

( ! ) Notice: Trying to get property of non-object in /app/public/wp-content/plugins/affiliate-wp/includes/misc-functions.php on line 993

DrewAPicture:
Oh yeah, it's cause we haven't loaded the upgrader at this point. I'll look into it.

DrewAPicture:
@pippinsplugins Try again after 2d27265e9fc25d68b509379fcb191ee680668886?

pippinsplugins:
Nope, still fatals.

DrewAPicture:
@pippinsplugins OK, give it another shot, please.

[...]

PR #2181, opened 2017-04-23
issue/2101 – Introduce a state of expiration for upgrades pinned to a specific version

DrewAPicture:
Issue: #2101


DrewAPicture:
Fool's errand. We'll just mark past routines complete on install. See #2242.

PR #2186, opened 2017-04-25
issue/2188 – Double escaping

tivnet:
Issue: #2188

Duplicate `esc_html` ?


DrewAPicture:
Definitely seems like a duplicate. Nice catch!

@tivnet One thing, can you change the additional return type to `false` instead of `bool`? We like to be as specific as possible :-)

tivnet:
@DrewAPicture Done.

DrewAPicture:
Awesome. We'll merge this as part of the 2.1 release. Thanks for the PR!

DrewAPicture:
Created #2188 for posterity.

PR #2236, opened 2017-05-28
Update affiliate-wp-es_ES.po

fernandotellado:
Minor changes to translation and solve fuzzy

Issue: #


DrewAPicture:
The fuzzy translations are still marked as such, which will continue to cause them not to be translated because they'll be ignored when the mo files are generated. Please remove the fuzzy or "needs work" notations.

fernandotellado:
Ah, ok. I usually use PoEdit but it doesn't works fine with github so I translate it directly. 

I'm going to delete the fuzzy notations and I'll send a pull request when finished

michaelbeil:
I am going to close this in favor of #2361.

Issue #2261, opened 2017-06-08
Null insert error if no description specified in WP Simple Pay Pro 2.5 (Stripe) shortcode.


pderksen:
The description attribute in the Stripe Checkout forms produced by WP Simple Pay Pro 2.5 is optional. In fact, all attributes are for one-time (non-subscriptions) payment forms.

If it's left out you'll get a null insert error.

WordPress database error: [Column 'description' cannot be null]
INSERT INTO `b5c_affiliate_wp_referrals` (`affiliate_id`, `date`, `currency`, `status`, `amount`, `reference`, `description`, `campaign`, `visit_id`, `products`, `custom`, `context`) VALUES (1, '2017-06-08 17:26:59', 'USD', 'pending', '0.20', 'ch_1ASUPPLwNybmtfEYzXS6r6IX', NULL, '', 1, '', 's:25:\"a:1:{s:8:\"livemode\";b:0;}\";', 'stripe')

Test WP Simple Pay shortcode:

`[stripe amount="100"]`

pippinsplugins:
Thanks @pderksen, we'll get this PR from @nickyoung87 merged in!

pippinsplugins:
PR looks good, merged!

DrewAPicture:
Backported to `master` via 0763e76c509b841bc0bba9e02d01acdc5e20b6e0. Milestone adjusted to 2.0.10.

PR #2266, opened 2017-06-09
#2261 Compatibility fix with WP Simple Pay version 3.0.0

nickyoung87:
Issue: #2261

Part of this fix is the empty description fix in the issue above. There is another part (no GH issue). The second part fixes the check for the Stripe email adress that gets posted. This is changed in WP Simple Pay 3.0.0 and I have added a check for that new value as well as keeping the old value in place for backwards compatibility.


michaelbeil:
Great work @nickyoung87!

PR #2283, opened 2017-06-15
Add affiliate object to the 'affwp_affiliate_deleted' hook

clickstudio:
Add an action that gets fired after the affiliate's data is removed.

It would have been be possible to use the 'affwp_affiliate_deleted' action but only the affiliate ID is passed which means I can't access the user ID (affiliate deleted).

Issue: #


clickstudio:
Perfect @DrewAPicture !!

DrewAPicture:
@clickstudio You're going to have to make the change because it's your fork. Just add the `$affiliate` arg and a parameter notation to the hook doc. The type is `\AffWP\Affiliate`.

clickstudio:
oh yeah, sorry @DrewAPicture 

Just made the changes!

PR #2365, opened 2017-09-04
Fix bad \r

fxbenard:
A pot issue


michaelbeil:
@fxbenard what are the specific issues you are seeing?

includes/admin/settings/class-settings.php
```diff
@@ -769,8 +769,7 @@ function get_registered_settings() {
 					),
 					'betas' => array(
 						'name' => __( 'Opt into Beta Versions', 'affiliate-wp' ),
-						'desc' => __( 'Receive update notifications for beta releases. When beta versions are available, an update notification will be shown on your Plugins page.
-', 'affiliate-wp' ),
```
fxbenard: @michaelbeil Here, a bad return. It breaks the pot and all po updates based on it.

michaelbeil:
Oh, okay. Can you open a new PR, without all of these extra commits, to fix this issue?

PR #2367, opened 2017-09-06
issue/2368 - Garbage collect in payout generation only if it is the first step

davefx:
If not, the whole data will be retrieved for every step in the payout generation

Issue: #2368


DrewAPicture:
Hi @davefx, thanks for the report. the `init()` method in all batch processes is _supposed_ to only run on the first step sp this check shouldn't be necessary. If `init()` is running on every step, then we should look at that as the source of the issue.

DrewAPicture:
Actually, I have that wrong. The logic in `pre_fetch()` should only run on the first step. So yeah, we're gonna have to fix all of the trash collections for all of the batch processes.

DrewAPicture:
This particular aspect has been fixed in #2371. Thanks for the report!

Issue #948, opened 2015-11-05
Improve Emails screen layout


lisag:
From a UX perspective, it would be awesome to have an **Email Options** section at the top of the _Emails_ screen for the general email options, and then give each email its own section (in case additional features are added to each specific email in the future). Perhaps just a simple divider line like in the General tab could make look a lot cleaner and easier to read. What do you guys reckon?

pippinsplugins:
Big =1

codelion7:
Definitely

pippinsplugins:
I agree.

gspice:
Love this suggestion, was just coming to create issue for something similar - can we bring all email content to this main area?  Thinking of the ones sent via code.

ramiabraham:
@AffiliateWP/core Added a PR - https://github.com/AffiliateWP/AffiliateWP/pull/2378

pippinsplugins:
@ramiabraham I think the `Notify Admin` option should be in the `Registration Email Options` section instead of above it.

ramiabraham:
@pippinsplugins Updated in https://github.com/AffiliateWP/AffiliateWP/commit/3c1012b56cd83c2ed488b24c1611db67946091d3

pippinsplugins:
@ramiabraham looks like we have a merge conflict now. Could you update it please?

ramiabraham:
@pippinsplugins @amdrew Merge conflict resolved in https://github.com/AffiliateWP/AffiliateWP/pull/2378/commits/d3ec81d738d71eddba88de39ff789cf4d8211b56

amdrew:
@ramiabraham 👍 Needs the same treatment for the other emails that show when require approval is enabled from the misc settings

ramiabraham:
@amdrew In `affwp_notify_on_registration`, the only check to the `require_approval` setting is a statement that appends a link in the admin email to review all pending applications, if `require_approval` is enabled:

```php
if( affiliate_wp()->settings->get( 'require_approval' ) ) {
    $message .= sprintf( __( 'Review pending applications: %s', 'affiliate-wp' ), affwp_admin_url( 'affiliates', array( 'status' => 'pending' ) ) ) . "\n\n";
}
```

All of the other emails are notifications for affiliates. What should be added to those?

pippinsplugins:
I don't think anything needs to be added to those.

amdrew:
[...]

PR #2378, opened 2017-09-07
issue/948 - Improve emails screen layout

ramiabraham:
- Adds clear separations between general email options and content settings for each email template.
- #948


pippinsplugins:
@ramiabraham this PR needs updated to resolve the merge conflict.

ramiabraham:
@pippinsplugins Updated

pippinsplugins:
@ramiabraham sorry, new conflict here now.

includes/emails/class-affwp-emails.php
```diff
@@ -532,13 +532,21 @@ public function email_tag_exists( $tag ) {
...
 	 */
 	public function is_email_disabled() {
-		$disabled = affiliate_wp()->settings->get( 'disable_all_emails', false );
-		$disabled = (bool) apply_filters( 'affwp_disable_all_emails', $disabled );
 
-		return $disabled;
+		$disabled = false;
+
+		// Emails are deemed disabled if no notifications are enabled.
+		if ( empty( affwp_get_enabled_email_notifications() ) ) {
```
pippinsplugins: This will trigger a fatal error on PHP 5.5 and earlier. `affwp_get_enabled_email_notifications()` needs to be set to a variable and then that variable passed to `empty()`.

includes/install.php
```diff
@@ -27,24 +27,41 @@ function affiliate_wp_install() {
...
-				'post_type'      => 'page',
-				'comment_status' => 'closed'
-			)
-		);
+
+		// Get the page ID of the Affiliate Area. 
+		$affiliates_page = affiliate_wp()->settings->get( 'affiliates_page' );
+
+		// Check that the page exists.
+		$affiliates_page = isset( $affiliates_page ) ? get_post( $affiliates_page ) : false;
```
pippinsplugins: `isset()` will always return `true` here, even if the `affiliates_page` setting is empty as `isset()` is checking that the variable itself is set to _something_, not that its set and truthful. To ensure proper behavior, change `isset()` to `! empty()`.

templates/dashboard-tab-settings.php
```diff
@@ -16,11 +16,13 @@
 			<input id="affwp-payment-email" type="email" name="payment_email" value="<?php echo esc_attr( $payment_email ); ?>" />
 		</div>
 
+		<?php if ( true === affwp_email_referral_notifications( absint( $affiliate_id ) ) ) : ?>
```
pippinsplugins: Let's remove `true ===` here. It's not really needed and it over complicates the readability of the template file without any benefit.

amdrew: Fair call 👍

includes/install.php
amdrew: Oops, copy/paste code from EDD's install script

includes/emails/class-affwp-emails.php
amdrew: Good catch!

Issue #1960, opened 2017-01-27
Referral Area - Settings: request for affiliate area setting to enable/disable New Referral Notifications per affiliate


gspice:
Requested by a few users so each affiliate doesn't have to go to Settings and check the box:  "Enable New Referral Notifications".

wordmaid:
This is more and more frequently requested in support. Have just had another one today. It would be awesome to see this implemented soon if possible.

When this option is globally enabled from the admin, the affiliate should then be able to go to their _Settings_ tab in the Affiliate Area and simply uncheck the option, and click `Save Profile Settings` to no longer receive **New Referral Notification** emails.

ckaklamanos:
This is a feature that I would like to see in AffialiateWP as well. It would be great to be able to make this a global setting. This is because when this is needed for all affiliates, admins should ask them explicitly to make this enabled in their Affiliate Area.

ramiabraham:
Related #1923

michaelbeil:
I believe so.

gspice:
I don't believe this is related to #1923 , rather, this is referring to the frontend  _Affiliate Area > Settings_ tab where affiliates have to opt-in to receive email notifications that a referral has been created for them.   Also, there are some privacy issues with this, meaning the admin would be opt-ing in for the affiliate on their behalf.

pippinsplugins:
@ramiabraham No it is not. What @gspice just said is correct.

ramiabraham:
@AffiliateWP/core PR ready for review https://github.com/AffiliateWP/AffiliateWP/pull/2379

michaelbeil:
Ticket: https://secure.helpscout.net/conversation/467597177/0/?folderId=205529.

michaelbeil:
https://secure.helpscout.net/conversation/483973806/72220?folderId=634609

pippinsplugins:
Attached PR is no longer valid. Will need to be re-done differently.

amdrew:
Working on this issue over in https://github.com/AffiliateWP/AffiliateWP/issues/948. 

All email notifications are now listed in a checkbox group, including the global option for allowing new referral notifications for affiliates. When enabled, the affiliate will see the option to enable new referral notifications in their affiliate area. They can also disable the email if they choose to, unless the admin deselects the global setting. In this instance, the checkbox option is removed from the affiliate area altogether.

Finally, all new affiliate registrations going forward will have new referral notifications enabled by default in the affiliate area.

[...]

PR #2379, opened 2017-09-08
issue/1960 - Add global referral notifications option

ramiabraham:
- Adds global `enable_default_new_referral_emails `, which is used to determine whether new referral email notifications are sent, if the user doesn’t have an `affwp_referral_notifications` meta key already set.
- Adds related check in `affwp_notify_on_new_referral`
- #1960

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ php:
     - 5.5
     - 5.6
     - 7.0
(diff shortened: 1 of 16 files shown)


pippinsplugins:
@ramiabraham could you update the PR to resolve the merge conflict?

Issue #1693, opened 2016-10-05
Affiliate area tab improvements


amdrew:
We should add a function that holds all the current tabs in AffiliateWP. This would allow https://github.com/AffiliateWP/affiliatewp-affiliate-area-tabs/issues/36 to be easily achievable.

In addition, other plugins should be able to add their tab to this function's array via a filter. This would allow the tab to show inside the Affiliate Area Tabs add-on settings. 

When re-ordering is possible (https://github.com/AffiliateWP/affiliatewp-affiliate-area-tabs/issues/8) this will be extremely handy, plus the user can deactivate the added tab/s from these plugins on demand.

As an example I want to add a new tab inside of Direct Link Tracking. I **can't** do the same as our other tabs on `dashboard.php`, added to the `affwp_affiliate_dashboard_tabs` hook:

```
<?php if ( affwp_affiliate_area_show_tab( 'direct-links' ) ) : ?>
<li class="affwp-affiliate-dashboard-tab<?php echo $active_tab == 'direct-links' ? ' active' : ''; ?>">
    <a href="<?php echo esc_url( affwp_get_affiliate_area_page_url( 'direct-links' ) ); ?>"><?php _e( 'Direct Links', 'affiliate-wp' ); ?></a>
</li>
<?php endif; ?>
```

Instead I have to do: 

```
<li class="affwp-affiliate-dashboard-tab<?php echo $active_tab == 'direct-links' ? ' active' : ''; ?>">
    <a href="<?php echo esc_url( add_query_arg( 'tab', 'direct-links' ) ); ?>"><?php _e( 'Direct Links', 'affiliate-wp' ); ?></a>
</li>
```

This is because there's no way to add `direct-links` to the following array and so the tab doesn't get it's `?tab=direct-links` query string:
https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/affiliate-functions.php#L1318

All in all, I feel we can clean our tabs up a bit to make it easier for other plugins to add their own tabs, but at the same time integrate these tabs with Affiliate Area Tabs, which is one of our most popular free add-ons.

ramiabraham:
@AffiliateWP/core PR ready for review: https://github.com/AffiliateWP/AffiliateWP/pull/2386

pippinsplugins:
@ramiabraham initial PR looks good.

Could you please test this with the `Affiliate Area Tabs` add-on to confirm everything still works properly with that?

ramiabraham:
[...]

PR #2386, opened 2017-09-12
issue/1693 - Affiliate area tab improvements 

ramiabraham:
- Introduces `affwp_affiliate_dashboard_tabs`, which gets all affiliate dashboard tabs, and returns a filterable array, `affwp_affiliate_dashboard_tabs`.
- Replaces hardcoded affiliate dashboard tab markup with a loop through `affwp_affiliate_dashboard_tabs`.
- #1693

### DIFF
--- a/includes/affiliate-functions.php
+++ b/includes/affiliate-functions.php
@@ -1555,12 +1555,19 @@ function affwp_get_lowest_priority_tab() {
 }
 
 /**
+ * /**
  * Returns all affiliate dashboard tabs in a filterable array.
  *
- * @return array  $tabs Affiliate dashboard tabs.
+ * @param  boolean $all     Whether to return all tabs, or only those which are visible.
+ *                          Default is true, which returns all tabs. Optional.
+ *
+ * @param  string  $exclude A dashboard tab ID, or array of IDs. If provided, these tabs will be excluded
+ *                          from the return value. Optional.
+ *
+ * @return array  $tabs     Affiliate dashboard tabs.
  * @since  2.1.7
  */
-function affwp_get_affiliate_dashboard_tabs( $exclude = '' ) {
+function affwp_get_affiliate_dashboard_tabs( $all = true, $exclude = '' ) {
 
 	/**
 	 * Filters the affiliate dashboard tabs.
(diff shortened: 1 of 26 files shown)


DrewAPicture:
Is this going to affect the Affiliate Area Tabs add-on at all?

templates/dashboard.php
```diff
@@ -53,16 +53,24 @@
 		<ul id="affwp-affiliate-dashboard-tabs">
 			<?php
 
-			$tabs = affwp_affiliate_dashboard_tabs();
+			$tabs = affwp_get_affiliate_dashboard_tabs();
+
+			// If the Affiliate Area Tabs add-on is hiding this tab, unset it from the $tabs array.
+			if ( class_exists( 'AffiliateWP_Affiliate_Area_Tabs' ) ) {
```
pippinsplugins: I don't believe this should be necessary. Affiliate Area Tabs uses the `affwp_affiliate_area_show_tab` filter to hide/show tabs. Am I wrong?

ramiabraham: It uses its own setting, `affiliate_area_hide_tabs`, which is checked here. If an affiliate area page is found, it's `unset` from the array.

pippinsplugins: @ramiabraham can we adjust it so the logic is purely in the add-on? As a general rule, we try to not put any add-on-specific code in core. The add-ons should extend and modify core, not the other way around.

ramiabraham:
@amdrew @pippinsplugins 
With AAT compatibility removed in https://github.com/AffiliateWP/AffiliateWP/pull/2386/commits/d53d2fa81bb85fb4856a99c5142cdc65e408401b, [this issue](https://github.com/AffiliateWP/affiliatewp-affiliate-area-tabs/issues/47) in AAT would need to be resolved in AAT prior to the 2.1.6 release (asking for feedback in review above).

pippinsplugins:
@ramiabraham looks like we have a merge conflict now.

includes/affiliate-functions.php
```diff
@@ -1506,6 +1506,120 @@ function affwp_get_active_affiliate_area_tab() {
...
+		}
+	}
+
+	/**
+	 * Filters the affiliate dashboard tabs.
+	 * Used to add, remove, or modify affiliate dashboard tabs.
+	 *
+	 * To add a new tab, provide an array containing the following:
+	 *
+	 *   $tabs['unique_tab_id'] => array(
```
amdrew: @ramiabraham Can we change `=>` to `=`

ramiabraham: Updated!

includes/affiliate-functions.php
```diff
@@ -1474,35 +1472,235 @@ function affwp_get_affiliate_area_page_url( $tab = '' ) {
...
+	/**
+	 * Filters the affiliate dashboard tabs.
+	 * Used to add, remove, or modify affiliate dashboard tabs.
+	 *
+	 * To add a new tab, provide an array containing the following:
+	 *
+	 *   $tabs['id'] = array(
+	 *      'title'    => __( 'Settings', 'affiliate-wp' ),
+	 *		'content'  => '',
+	 *		'priority' => 10
```
amdrew: Indentation on content and priority needs tweaking

PR #2387, opened 2017-09-13
Add option to set cookies for sub-domains

maxrice:
This PR expands on #2140/#2142 to:

- add a setting to allow sharing tracking cookies with sub-domains (disabled by default)
screen shot 2017-09-13 at 4 59 38 pm

- automatically use the WP `COOKIE_DOMAIN` constant, if set

This could be considered a breaking change since previously the `COOKIE_DOMAIN` was only used for fallback (server-side) tracking, but not JS tracking.

### DIFF
--- a/assets/js/tracking.js
+++ b/assets/js/tracking.js
@@ -368,12 +368,18 @@ jQuery(document).ready( function($) {
 	}
 
 	/**
-	 * Set a cookie, with optional domain if set.
+	 * Set a cookie, with optional domain if set. Note that providing *any* domain will
+	 * set the cookie domain with a leading dot, indicating it should be sent to sub-domains.
 	 *
-	 * @since
+	 * example: host.tld
 	 *
-	 * @param {string} name
-	 * @param {string} value
+	 * - $.cookie( 'some_cookie', ...) = cookie domain: host.tld
+	 * - $.cookie ('some_cookie', ... domain: 'host.tld' ) = .host.tld
+	 *
+	 * @since 2.x.x
+	 *
+	 * @param {string} name cookie name, e.g. affwp_ref
+	 * @param {string} value cookie value
 	 */
 	function affwp_set_cookie( name, value ) {
 
--- a/assets/js/tracking.js
+++ b/assets/js/tracking.js
@@ -341,7 +341,7 @@ jQuery(document).ready( function($) {
 	function affwp_track_visit( affiliate_id, url_campaign ) {
 
 		// Set the cookie and expire it after 24 hours
-		$.cookie( 'affwp_ref', affiliate_id, { expires: AFFWP.expiration, path: '/' } );
+		affwp_set_cookie( 'affwp_ref', affiliate_id );
 
 		// Fire an ajax request to log the hit
 		$.ajax({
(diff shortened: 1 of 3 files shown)


includes/class-tracking.php
```diff
@@ -747,7 +751,23 @@ public function ajax_get_affiliate_id_from_login() {
...
+	/**
+	 * Get the cookie domain
+	 * @return bool|string
+	 */
+	public function get_cookie_domain() {
+
+		// COOKIE_DOMAIN is false by default
+		$cookie_domain = COOKIE_DOMAIN;
+
+		if ( ! $cookie_domain && affiliate_wp()->settings->get( 'cookie_sharing', false ) ) {
```
DrewAPicture: Let's split out the second conditional to a separate initialization. Other you have to do a little head-tilting to grasp the "if true is false" logic.

includes/class-tracking.php
```diff
@@ -747,7 +751,23 @@ public function ajax_get_affiliate_id_from_login() {
 	 * @since 1.0
 	 */
 	public function set_affiliate_id( $affiliate_id = 0 ) {
-		setcookie( 'affwp_ref', $affiliate_id, strtotime( '+' . $this->get_expiration_time() . ' days' ), COOKIEPATH, COOKIE_DOMAIN );
+		setcookie( 'affwp_ref', $affiliate_id, strtotime( '+' . $this->get_expiration_time() . ' days' ), COOKIEPATH, $this->get_cookie_domain() );
+	}
+
+	/**
+	 * Get the cookie domain
+	 * @return bool|string
```
DrewAPicture: Needs a full return description and a period on the summary.

assets/js/tracking.js
```diff
@@ -368,6 +368,23 @@ jQuery(document).ready( function($) {
 	}
 
 	/**
+	 * Set a cookie, with optional domain if set.
+	 *
+	 * @since
+	 *
+	 * @param {string} name
```
DrewAPicture: Let's get full parameter descriptions here. Go ahead and set the `@since` to `2.x.x`.

DrewAPicture:
Added some requested changes, but otherwise this seems to work as expected for me. Nice work.

assets/js/tracking.js
ramiabraham: `console.affwp()` can be used to here to save the extra calls to `window` objects.

maxrice: >console.affwp() can be used to here to save the extra calls to window objects.

not sure I follow this one, could you give an example of how to use that here?

maxrice:
thanks for the review & feedback @DrewAPicture @ramiabraham -- I've added changes in a0c650a30f7eb841f88f25758bc57fc5b7986aac. Any other changes?

DrewAPicture:
@ramiabraham Can you please review the latest changes and mark approved if it looks good?

maxrice:
hey all, any further changes needed for this? would love to be able to update to the latest version when this PR is included 😸

pippinsplugins:
@maxrice apologies for the delay here! We'll get this included as soon as possible.

maxrice:
just wanted to surface this PR again 😄

pippinsplugins:
@maxrice merged finally! :)

maxrice:
thanks!

Issue #2393, opened 2017-09-23
AffiliateWP metabox does not show on a Give donation form 


michaelbeil:
When testing Give locally, I don't see the AffiliateWP metabox when I go to edit the donation form. It appears like we need to update our Give integration.

DrewAPicture:
Looks like this has been broken since Give 1.8. They [deprecated use of CMB2](https://givewp.com/give-version-1-8/) in favor of their own settings API. Example here: https://github.com/WordImpress/Give-Settings-API-Example

michaelbeil:
Should we release a fix for this in an upcoming release?

pippinsplugins:
Yes let's go ahead and fix this for 2.1.6.

ramiabraham:
Added a PR: https://github.com/AffiliateWP/AffiliateWP/pull/2400

Presently finishing up the rate calc. Can we add a param for the rate in the base class `calculate_referral_amount` method? It'd add a param for `$rate`, then just a ternary or something:

```php
if ( ! empty( $product_id ) ) {
			$rate = ! empty( $rate ) ? $rate : $this->get_product_rate( $product_id, $args = array( 'reference' => $reference, 'affiliate_id' => $affiliate_id ) );
		}
```

pippinsplugins:
@ramiabraham one minor change needed. See review on PR>

pippinsplugins:
@amdrew could you give this a test today?

michaelbeil:
This PR fixes the metabox box issue for me.

amdrew:
@pippinsplugins Works for me too

PR #2400, opened 2017-10-03
issue/2393 - Update Give integration to use new API

ramiabraham:
- Updates Give integration to use new API
- #2393 - Pending addition of custom rate calculation (in progress)

### DIFF
--- a/includes/integrations/class-give.php
+++ b/includes/integrations/class-give.php
@@ -47,8 +47,8 @@ class Affiliate_WP_Give extends Affiliate_WP_Base {
 		$affiliate_id = $this->get_affiliate_id( $payment_id );
 
 		// Referral rate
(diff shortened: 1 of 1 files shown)


includes/integrations/class-give.php
```diff
@@ -58,7 +62,7 @@ public function add_pending_referral( $payment_id = 0, $payment_data = array() )
 		}
 
 		// Get referral total
-		$referral_total = $this->get_referral_total( $payment_id, $affiliate_id );
+		$referral_total = affwp_calc_referral_amount( $referral_total, $affiliate_id, $payment_id, $rate, 0 );
```
pippinsplugins: `$referral_total` has not been defined yet so it cannot be passed to `affwp_calc_referral_amount()`

ramiabraham: Corrected!

Issue #2273, opened 2017-06-11
Remove dependency on Views


pippinsplugins:
Unfortunately it seems the time has come to ditch our dependence on MySQL Views for the campaigns table.

See #1688 and #1074 for related issues.

We have tried to work with hosts to get them to properly implement support for MySQL Views, which _should_ be a standard feature but, sadly, it seems this is unlikely to happen in any reasonable timeframe. While we continue to wait, customers are negatively affected in a way that they never should be.

The quick summary of the problem is that hosts and backup / migration plugins that offer backup and restore functionality oftentimes do not work with views. Not only do they not work but the presence of a view causes the entire backup / restore process to fail.

Here's an [example ticket](https://secure.helpscout.net/conversation/379419613/60860/?folderId=205530) of a customer that is simply unable to use AffiliateWP unless he:

- Moves to a different host (currently on WPEngine)
- Does not use WPEngine's backup system
or
- WPengine adds proper support for views.

This is not the first, nor even the 10th, ticket we've received related to this problem.

For those that cannot view the ticket linked above, here's an excerpt from it:

>>>A) On WP Engine my database user has ALL PRIVELEGES on my database and the import still fails.
B) I decided to try using a different host: LiquidWeb's managed WordPress hosting. With them my db user also has ALL PRIVILEGES for my database. Also with them, I am not able to restore from backup if AffiliateWP is enabled.
C) On my local machine using XAMPP my db user has ALL PRIVILEGES and I can clearly see that I have the SHOW VIEW privilege but exporting & then importing the database fails.

>>>So I did some research into this and I believe the problem is this bug in mysqldump:

>>>https://bugs.mysql.com/bug.php?id=24680

>>>If I remove the DEFINER line from the sql file then the import completes successfully on my local machine.

>>>/*!50013 DEFINER=`awptest2`@`127.0.0.1` SQL SECURITY DEFINER */

>>>Unfortunately, I do not have the necessary access to modify the sql file on WP Engine or LiquidWeb before the the import takes place and I would imagine most of your customers who are on shared hosting do not have this access either. There are some workarounds that involve writing a script to automatically strip this DEFINER line from from mysqldump files:

>>>https://dbperf.wordpress.com/2010/04/12/removing-definer-from-mysql-dump/
>>>https://stackoverflow.com/questions/9446783/remove-definer-clause-from-mysql-dumps

>>>..but of course I do not have access to implement this in a shared hosting environment.

When we create the view, we simply do this:

[...]

PR #2406, opened 2017-10-06
Issue/2273 - Remove dependency on Views

ramiabraham:
- Remove dependency on Views
- WIP
- #2273

### DIFF
--- a/includes/admin/tools/upgrades/class-batch-upgrade-remove-views.php
+++ b/includes/admin/tools/upgrades/class-batch-upgrade-remove-views.php
@@ -1,0 +1,143 @@
+<?php
+namespace AffWP\Utils\Batch_Process;
+
+use AffWP\Utils;
+use AffWP\Utils\Batch_Process as Batch;
+
+/**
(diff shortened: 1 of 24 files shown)


includes/class-campaigns-db.php
```diff
@@ -276,60 +351,85 @@ public function count( $args = array() ) {
...
+
 	}
 
 	/**
-	 * Create the view
+	 * Creates the table.
 	 *
-	 * @since  1.7
-	 */
-	public function create_view() {
```
DrewAPicture: Can't just rename it. We need to properly deprecate it :-)

includes/admin/tools/upgrades/class-batch-upgrade-remove-views.php
```diff
@@ -0,0 +1,143 @@
...
+
+
+
+	/**
+	 * Defines logic to execute once batch processing is complete.
+	 *
+	 * @access public
+	 * @since  2.1.6
+	 */
+	public function finish() {
```
DrewAPicture: Note that as of 2.1.6 there's a required parameter for `finish()` and `parent::finish()`.

PR #2425, opened 2017-10-30
 issue/2426 – Adding note to order also when updating referral

davefx:
When updating an existing referral, if its original amount and/or affiliate changes, write a note to the order.

Issue: #2426


michaelbeil:
Resolved in https://github.com/AffiliateWP/AffiliateWP/pull/2482.

Issue #2450, opened 2017-11-09
Visits graph shows incorrect totals when filtered by an affiliate


gspice:
PR: #2456

Reported on this ticket:
https://secure.helpscout.net/conversation/463865185/70033?folderId=1170842

When you view the Visits graph for an affiliate, the numbers on the graph are incorrect.  The all time visits (as of today) were 70, but the graph shows 2563 unconverted and 186 converted when hovering over the graph point.  

<img width="1372" alt="affiliate-visits-graph" src="https://user-images.githubusercontent.com/3018245/32581243-46d9e622-c4a6-11e7-9a04-63c4e36dd49d.png">

DrewAPicture:
After some investigation, it looks like the root of the problem lies in query for date-based objects throughout core without first converting dates to UTC (essentially, we're currently querying based on "now" dates instead of first accounting for the GTM offset in incomine date values). This appears to be widespread and reaches further than just visit queries. Working on a PR + tests.

DrewAPicture:
Started a WIP PR: #2454

DrewAPicture:
Closed #2454 and started a new WIP PR with a different approach: #2456

DrewAPicture:
The attached PR, #2456 is ready for testing and review:

@gspice @amdrew 

Testers:

The primary things to test with this PR are the following:
1. Filtering reports by specific dates or "today" should reflect expected results
2. Any dates and times showing in AffWP admin screens should now be displaying in a nicer format and in "WordPress time" – that is, respecting the WordPress timezone settings for display. If you haven't yet set the Timezone for your test site in General > Settings, you should do that.

The primary bug the PR fixes is that reports weren't always reflecting objects created for the time period being filtered for. It turns out that this was because we weren't translating dates/date ranges entered in the UI into UTC times to ensure that objects returned matched those dates. We store the dates/times in database in UTC (GMT+0) to ensure a "neutral" time scheme, this allows the site to adjust the timezone as needed without screwing up the actual data that's being stored.

The PR not only fixes querying for objects with WordPress dates/times to return UTC-date/time-based results, but adjusting all admin and template views to display dates/times in "local" or "WordPress" time.

For example, let's say you have a referral created at 23:00 UTC, if your Timezone is set to Denver (-7), the time would be _displayed_ as 4:00 pm. An easy way to confirm times and dates are showing in "WordPress time" is to retrieve the object you're checking in the command line first:

[...]

PR #2454, opened 2017-11-13
Issue/2450 – Account for GMT offset when querying for objects by date (WIP)

DrewAPicture:
Issue: #2450


DrewAPicture:
Closed in favor of a new PR with a slightly different approach.

Issue #2474, opened 2017-12-05
Cannot find filter to avoid payout generation


davefx:
In the site I'm working on, my client has asked me to avoid the generation of payouts for inactive affiliates. 
I already have located filters that allow me to avoid the generation of referrals for these affiliates, but I cannot find any filter to avoid the generation of payouts once the referrals have been created.

DrewAPicture:
After some back and forth on the original PR, I've submitted an alternative: #2479, which introduces a new dynamic `affwp_pre_insert_$type_data` filter in `db->insert()`. Passing a falsey value back via a filter callback will cause `insert()` to bail. Includes 5 tests on a variety of falsey values.

```php
add_filter( 'affwp_pre_insert_payout_data', '__return_false' );
```

PR #2475, opened 2017-12-05
issue/2474 – Adding new filter to filter args when adding a payout

davefx:
Issue: #2474


includes/class-payouts-db.php
```diff
@@ -165,6 +165,15 @@ public function add( $args = array() ) {
 			$args['referrals'] = (array) absint( $args['referrals'] );
 		}
 
+		/**
+		* Filters the whole arguments when adding a payout.
```
DrewAPicture: Looks like there's some weirdness with the indentation. There should be a leading space before the asterisks in each line of the DocBlock.

includes/class-payouts-db.php
```diff
@@ -165,6 +165,15 @@ public function add( $args = array() ) {
...
+		*
+		* @since 2.2
+		*
+		* @param array  $args      Data for adding a payout.
```
DrewAPicture: It looks like you've tabbed apart the elements on the `@param` line, these can just be spaced apart with a single space between each element.

includes/class-payouts-db.php
DrewAPicture: This could be more descriptive, and I'd prefer that we move the filter down a few lines to just before where the payout is actually inserted into the database. Just before this line: `$add = $this->insert( $args, 'payout' );`

The way we typically add context in the summary is something along the lines of "Filters the arguments used to create a payout just before it is inserted into the database."

davefx: The important point is that we want to provide a way to avoid the payout insertion. 
We don't need to change the generated payout, but completely prevent its generation in some cases.

davefx: Fixed.

includes/class-payouts-db.php
davefx: Fixed

DrewAPicture:
If the entire point is that you just want to prevent the payout, then a filter on the args to vicariously accomplish that is not how I would prefer to go about it. The better option would be to put in a filter _specifically_ for bailing from `insert()`. As such, I've implemented such a filter in PR: #2479.

I'm going to close this PR in favor of that.

Issue #2490, opened 2017-12-30
Identify the Affiliate Area page in the list table screen


amdrew:
This would also be super handy for showing the `Affiliate Area` page: https://github.com/restrictcontentpro/restrict-content-pro/issues/1617

I find myself looking at the page set in the settings, then trying to find it in the pages list.

tubiz:
PR: #2494

PR #2494, opened 2018-01-03
issue/2490 - Identify the Affiliate Area page in the list table screen

tubiz:
Issue: #2490


affiliate-wp.php
```diff
@@ -382,6 +382,7 @@ private function includes() {
 			require_once AFFILIATEWP_PLUGIN_DIR . 'includes/admin/tools/class-migrate.php';
 			require_once AFFILIATEWP_PLUGIN_DIR . 'includes/admin/add-ons.php';
 			require_once AFFILIATEWP_PLUGIN_DIR . 'includes/admin/user-profile.php';
+			require_once AFFILIATEWP_PLUGIN_DIR . 'includes/admin/posts-list.php';
```
DrewAPicture: Rather than introduce a new top-level admin file just for this, let's put this code in includes/actions.php.

tubiz:
The requested changes has been made.

includes/actions.php
```diff
@@ -93,3 +93,22 @@ function affwp_process_add_affiliate_website( $affiliate_id, $args ) {
 
 }
 add_action( 'affwp_insert_affiliate', 'affwp_process_add_affiliate_website', 11, 2 );
+
+/**
+ * Add a post display state for AffiliateWP pages in the page list table.
```
DrewAPicture: Minor nitpick here, we need to use third-person singular verbs in DocBlock summaries. In this case, that would be "Adds". Everything else looks great.

tubiz: Updated. Thanks for pointing it out.

Issue #2411, opened 2017-10-17
WooCommerce:  add link in order detail to go directly to Referral edit screen


gspice:
From support ticket:
https://secure.helpscout.net/conversation/450483126/68651?folderId=1170842

>Not sure if this possible with Woocommerce order notes, but if you could make the referral number clickable to take you right into the referral edit screen that would be awesome.

>If that is not possible, maybe if somewhere in the order info near the top you could add link to edit the referral that would be helpful when that is needed to save copying it, going to the referrals section and then pasting it while remembering to enter the correct search variable like "ref:", etc.

pippinsplugins:
PR is ready for testing from @tubiz!

tubiz:
PR: #2498

![woocommerce order note](https://user-images.githubusercontent.com/602184/34546498-a5edcf96-f10d-11e7-9f28-7a998077149b.png)

pippinsplugins:
Works great, merged!

PR #2498, opened 2018-01-04
issue/2411 - WooCommerce: Add link in order note to go directly to referral edit screen

tubiz:
Issue: #2411

### DIFF
--- a/includes/integrations/class-woocommerce.php
+++ b/includes/integrations/class-woocommerce.php
@@ -216,10 +216,11 @@ class Affiliate_WP_WooCommerce extends Affiliate_WP_Base {
 
 					$this->log( sprintf( 'Referral #%d created successfully.', $referral_id ) );
 
-					$amount = affwp_currency_filter( affwp_format_amount( $amount ) );
-					$name   = affiliate_wp()->affiliates->get_affiliate_name( $affiliate_id );
+					$amount         = affwp_currency_filter( affwp_format_amount( $amount ) );
+					$name           = affiliate_wp()->affiliates->get_affiliate_name( $affiliate_id );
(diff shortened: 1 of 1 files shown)


includes/integrations/class-woocommerce.php
```diff
@@ -216,10 +216,11 @@ public function add_pending_referral( $order_id = 0 ) {
...
 					$this->log( sprintf( 'Referral #%d created successfully.', $referral_id ) );
 
-					$amount = affwp_currency_filter( affwp_format_amount( $amount ) );
-					$name   = affiliate_wp()->affiliates->get_affiliate_name( $affiliate_id );
+					$amount         = affwp_currency_filter( affwp_format_amount( $amount ) );
+					$name           = affiliate_wp()->affiliates->get_affiliate_name( $affiliate_id );
+					$referral_link  = esc_url( affwp_admin_url( 'referrals', array( 'referral_id' => $referral_id, 'action' => 'edit_referral' ) ) );
 
-					$this->order->add_order_note( sprintf( __( 'Referral #%d for %s recorded for %s', 'affiliate-wp' ), $referral_id, $amount, $name ) );
+					$this->order->add_order_note( sprintf( __( 'Referral <a href="%1$s">#%2$d</a> for %3$s recorded for %4$s', 'affiliate-wp' ), $referral_link, $referral_id, $amount, $name ) );
```
DrewAPicture: Let's use `affwp_admin_link()` and a single specifier in the note instead of `affwp_admin_url()`. This will allow you to keep HTML out of the translatable string and just pass `$referral_link` in as a single specifier. The same applies to the one 30 lines down too.

Also, let's add a translators note on the line before noting what all of the specifiers represent. See [here](https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/admin/referrals/edit.php#L53) for how to go about that.

tubiz: Updated.

Issue #2198, opened 2017-05-02
New referral type for opt-in/leads (and possibly others)


amdrew:
In our meeting we discussed generating referrals when a user "opts-in". This would generally occur when the visitor (aka lead) signs up to a mailing list. Rather than building this as an add-on (https://trello.com/c/XsqTj2PL/244-opt-in-referrals) we're exploring the possibility of including this in core since we already create referrals when a form is submitted, when a purchase is made etc.

For the above to happen, we should allow custom referral types to be created. This would also allow us to filter referrals and see _how_ referrals were generated. Was the referral generated from an eCommerce purchase? From a form submission? From an opt-in?

Here's another example of how custom referral types would help other developers:
https://github.com/AffiliateWP/AffiliateWP/issues/1100

An older related discussion:
https://github.com/AffiliateWP/AffiliateWP/issues/505

pippinsplugins:
When we do this, we can also then add integrations for plugins like these:

- OptinMonster
- MailChimp for WordPress
- Popup Builder
- Etc

DrewAPicture:
Changed "status/statuses" to "type/types". Statuses are statuses, types are types ;)

pippinsplugins:
I've added a new `type` column and set all existing referrals to the type of `sale` through an upgraded routine.

For the moment I have decided against requiring that referral types be formally registered (unlike post types in WP core). The primary reason for this is that we're not (at least for now) revealing anything through the UI for these referral types, so it's not necessary that the available types be strictly known.

pippinsplugins:
I've changed my mind on having a defined method for registering referral types.

We need to have a way to filter the Referrals table by type so that depends on having a registry where we can retrieve a list of all available types.

I've added a registry and added a filter dropdown to the referrals table.

pippinsplugins:
This is ready for testing.

By default we're now registering two core types:

- `sale`
- `opt-in`

New types can be registered like this:

```php
function pw_affwp_register_referral_type( $registry ) {
	$registry->register_type( 'ping', array( 'label' => 'Ping' ) );
}
add_action('affwp_referral_type_init', 'pw_affwp_register_referral_type' );
```

The `type` parameter is supported in the `add()`, `update()`, and `get_referrals()` methods.

The Referrals list table can be filtered by any registered type.

If a referral is created with a non-registered type, it will default to `sale`.

amdrew:
[...]

PR #2521, opened 2018-01-14
Add a type column to the referrals database. #2198

pippinsplugins:
#2198


includes/class-referral-type-registry.php
```diff
@@ -0,0 +1,121 @@
...
+	/**
+	 * Retrieves a type and its associated attributes.
+	 *
+	 * @access public
+	 * @since  2.2
+	 *
+	 * @param string $type_id Type ID.
+	 * @return array|false Array of attributes for the type if registered, otherwise false.
+	 */
+	public function get_type( $type_id ) {
```
DrewAPicture: You can skip defining methods that just wrap the parent methods and use magic methods instead. That's the method I'm using for new registries in the reports API, at least.

See [here](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/6273/files#diff-f4ffa2fff58716d72afbcd817bf5d6ebR23) for the `@method` notations. See [here](https://github.com/easydigitaldownloads/easy-digital-downloads/pull/6273/files#diff-f4ffa2fff58716d72afbcd817bf5d6ebR60) for the magic methods.

Just a thought.

pippinsplugins: Huh, that's interesting. Obviously both work, what do you think is the benefit to the magic methods?

DrewAPicture: Pretty much the only useful aspect of using the magic methods is not defining extra methods for the sole purpose of namespacing the parent functions. If there's a method, it should have a valid purpose, and I'm not sure I consider namespacing to be enough of a reason. Granted, I didn't do it this way previously in AffWP, but that's my thinking nowadays.

pippinsplugins: I think for consistency's sake we should either namespace all of them (like AFFWP currently does) or move them all to magic methods.

Issue #2303, opened 2017-07-06
Pagination does not work if Affiliate Area is set to be the "front page"


amdrew:
If the affiliate area is set to be the "front page" (Setting -> Reading) then the pagination on the visits or referrals tab does not work. Clicking the page numbers reloads the page and does not show the correct results.

pippinsplugins:
@tubiz Would be great if this could be tackled as well.

tubiz:
PR: #2542

PR #2542, opened 2018-01-30
Issue/2303 - Fix pagination does not work if Affiliate Area is set to be the "front page"

tubiz:
Issue: #2303


tubiz:
@amdrew PR updated.

pippinsplugins:
I approve of `affwp_get_current_page_number()`. Definitely helps simplify it.

tubiz:
I will add the test for the new function.

amdrew:
PR update looks good, can be merged after test added.

tubiz:
@DrewAPicture I am stuck in writing the test for the `affwp_get_current_page_number()`. Can you direct me?

DrewAPicture:
@tubiz Let me take a look.

includes/misc-functions.php
```diff
@@ -1233,3 +1233,20 @@ function affwp_maybe_unserialize( $original ) {
...
+/**
+ * Retrieves the current page number.
+ *
+ * @since 2.1.12
+ *
+ * @return int The current page number.
+ */
+function affwp_get_current_page_number() {
+	if ( is_front_page() ) {
+		$page = get_query_var( 'page' ) ? get_query_var( 'page' ) : 1;
```
DrewAPicture: `get_query_var()` accepts a second optional `$default` parameter, so you could simplify these two ternaries.

DrewAPicture:
@tubiz @amdrew Yeah, so this is actually super hard to test and not really possible to make it consistent. So we should probably just skip the test(s) since we're really just working around a quirk of WP core anyway.

amdrew:
@DrewAPicture 👍

includes/misc-functions.php
tubiz: I will update the PR shortly.

tubiz:
@amdrew @DrewAPicture PR updated. Thanks.

Issue #2540, opened 2018-01-29
Cookie/visit not being created with fallback tracking and usernames with periods


amdrew:
Similar to https://github.com/AffiliateWP/AffiliateWP/issues/2539

If Fallback Tracking is enabled, referral links with a period in the username do not generate a cookie/visit. Non-pretty referral links work however.

### Works and generates a visit
https://issues.affwp.dev/?ref=test.user

### Does not generate a visit
https://issues.affwp.dev/ref/test.user

### To replicate:

1. Enable `Use Fallback Referral Tracking Method` from _Affiliates > Settings > Misc_.
2. Create or use an affiliate's link who has a period in their username.
3. Use the pretty version of the affiliate link, e.g. `https://issues.affwp.dev/ref/test.user`. The cookie/visit will not be created.

Disabling fallback tracking allows the same referral link to store a cookie/visit.

Ticket:
https://secure.helpscout.net/conversation/512934442/74899?folderId=205529

amdrew:
@tubiz Could you please also tackle this one, similar to #2539 so possibly related.

tubiz:
PR: #2549

PR #2549, opened 2018-01-31
Cookie/visit not created if affiliate username is an email, contains a period when fallback tracking is enabled. Fixes #2539 #2540

### LINKED ISSUE #2539: Visit/cookie not created in browser if affiliate username is email address and fallback tracking is enabled.
Discovered on support ticket:
https://secure.helpscout.net/conversation/511477820/74698?folderId=205529

Setup:

* Enable **Use Fallback Referral Tracking Method** under _Affiliates > Settings > Misc_.
* Create or use an affiliate's link who has an email address as their username.
* Use the link as standard or "pretty" link in a browser, the cookie/visit will not be created.

Workaround:
* Turn off Fallback tracking and the cookie/visit is created successfully.

tubiz:
Issue: #2539 #2540

### DIFF
--- a/includes/class-tracking.php
+++ b/includes/class-tracking.php
@@ -546,8 +546,9 @@ class Affiliate_WP_Tracking {
 			if ( false !== strpos( $path, $this->get_referral_var() . '/' ) ) {
 
 				$pieces = explode( '/', str_replace( '?', '/', $path ) );
-				$pieces = array_map( 'sanitize_key', $pieces );
 				$key    = array_search( $this->get_referral_var(), $pieces );
+
+				$pieces[ $key + 1 ] = strtolower( sanitize_user( $pieces[ $key + 1 ] ) );
 
 				if ( $key ) {
 


tubiz:
I wasn't able to use `john+sitename@domain.com.` or `john+sitename@domain.com` as the username when creating an affiliate account from the affiliate registration page or when adding a new user from the admin and the 'Add the user as an affiliate' checkbox is checked as the username is validated using the `validate_username()` and will return false.

But I was able to use both username on the 'Add New Affiliate' page in the admin but the `+` (plus symbol) was removed from both usernames.  

The username need to be validated in the `affwp_process_add_affiliate()` to prevent username having a `+` (plus symbol) from being used. So we can still keep using `urldecode ()` https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/class-tracking.php#L701

amdrew:
@tubiz Ok then, let's leave it for now. I tested by updating the username directly in the DB, totally forgot our own forms did not allow it.

Issue #2527, opened 2018-01-20
PayPal Buttons JS error: TypeError: action is undefined


amdrew:
A customer's site is generating a JavaScript error when the PayPal Buttons integration is enabled. 

```
TypeError: action is undefined
```

This is appearing when a modal window is being launched after adding a WooCommerce product to the cart, provided by another WooCommerce related plugin: 

<img width="939" alt="screenshot 2018-01-20 14 41 06" src="https://user-images.githubusercontent.com/52581/35178684-46979022-fdf1-11e7-90f1-452516d1b52e.png">

Because of the JS error, the "no thanks" and `x` links cannot be clicked.

The specific line referenced in the HTML source is coming from our PayPal integration:

```
`if ( ! action.match( paypalMatch ) ) {`
```

Ticket: https://secure.helpscout.net/conversation/508945591/74456?folderId=205527

tubiz:
@pippinsplugins on it.

amdrew:
👍 This is causing another customer's checkout page to break: https://secure.helpscout.net/conversation/513545519/74968?folderId=205527

amdrew:
Another report. This time an "update your payment form".

tubiz:
@amdrew can you send the sites url so I can check.

amdrew:
@tubiz Here's how you can replicate the issue, based on the most recent report:

1. Install WooCommerce
2. Install this plugin: https://wordpress.org/plugins/woocommerce-gateway-paypal-powered-by-braintree/
3. Enable the `Braintree (Credit Card)`. 
4. Enable sandbox mode and enter any details. If you want, sign up to braintree and put in some real test API credentials. You'll then see a real form instead of the error shown below.

    <img width="678" alt="screenshot 2018-02-03 12 01 59" src="https://user-images.githubusercontent.com/52581/35758693-24d29c9a-08da-11e8-814b-11fdc2252f0f.png">
5. Go to `/my-account` on the front-end
6. You'll then see a `Payment Methods` section
    <img width="894" alt="screenshot 2018-02-03 12 03 39" src="https://user-images.githubusercontent.com/52581/35758753-5c757dca-08da-11e8-9d91-ee8bfc5fc169.png">
7. Click `Add Payment Method` and you'll see the JS error in the console. Make sure you have the PayPal Buttons integrations enabled in AffWP's settings to see this.
[...]

PR #2553, opened 2018-02-02
Fix issue with submitting forms without the "action" attribute

albionselimaj:
There's a typo when checking if the "action" attr is present in the form element
instead of `if ( 'undefined' === typeof action )`
it's `if ( 'undefined' === action )`

This breaks forms without an action, as it throws an error later when trying to match `action.match( paypalMatch )`

Issue: #2527


amdrew:
Thanks @albionselimaj! Good catch. I can verify that this fixes the issue 👍

Issue #2572, opened 2018-02-16
Hyphen being added to referral description for standard downloads


amdrew:
When purchasing a standard product in EDD (non-variable priced), the referral's description is adding a hyphen:

<img width="460" alt="screen shot 2018-02-16 at 1 30 29 pm" src="https://user-images.githubusercontent.com/52581/36288030-95d3da4c-131d-11e8-93b5-fa61d19218b4.png">

I'm wondering if something changed in EDD recently, since I do believe it used to work since we introduced the hyphen: https://github.com/AffiliateWP/AffiliateWP/issues/2380.

In any case, the `price_id` for a standard download is being reported as `(int) 0`, and for the first variable priced option of a download it's `(string) 1`. The condition is failing because it's not `null` and the hyphen is being added: https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/integrations/class-edd.php#L459

amdrew:
PR: https://github.com/AffiliateWP/AffiliateWP/pull/2573

Works fine now in my testing:

<img width="490" alt="screen shot 2018-02-16 at 1 44 03 pm" src="https://user-images.githubusercontent.com/52581/36288441-7ac5ec16-131f-11e8-82d9-65e00fcefed7.png">

PR #2573, opened 2018-02-16
Issue/2572 - Use ! empty()

amdrew:
#2572

### DIFF
--- a/includes/integrations/class-edd.php
+++ b/includes/integrations/class-edd.php
@@ -450,14 +450,17 @@ class Affiliate_WP_EDD extends Affiliate_WP_Base {
 
 		foreach ( $downloads as $key => $item ) {
 
-			if ( get_post_meta( $item['id'], '_affwp_' . $this->context . '_referrals_disabled', true ) ) {
+			$download_id = $item['id'];
+			$download    = new EDD_Download( $download_id );
+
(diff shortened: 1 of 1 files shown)


includes/integrations/class-edd.php
```diff
@@ -456,7 +456,7 @@ public function get_referral_description( $payment_id = 0 ) {
 
 			$desc = get_the_title( $item['id'] );
 
-			if ( isset( $item['options']['price_id'] ) && null !== $item['options']['price_id'] ) {
+			if ( ! empty( $item['options']['price_id'] ) ) {
```
pippinsplugins: This will fail for older products that have a price ID of `0`.

amdrew: Ahh yes, that's right, they used to be indexed at `0`. Got it

amdrew:
@pippinsplugins See latest commit, happy with that?
https://github.com/AffiliateWP/AffiliateWP/pull/2573/commits/eab072eee5d6018f28faceda5449922475824205

It seems like checking if a product is actually variable priced is a better option than checking against the `price_id`. Method is supported since EDD v2.2, customers should upgrade if on an earlier version of EDD.

pippinsplugins:
Yeah that's great, I like that.

Issue #2570, opened 2018-02-15
Admin:  Adding new referral manually doesn't display an error if the affiliate selection is not completed.


gspice:
When manually adding a new referral, if the Affiliate is not selected, when you click **Add Referral**, it appears to save and you are returned to the Referral list, but nothing has been saved.   Desired behavior is an error of some sort as displayed on the Add Affiliate screen when the User login name has not been found.

**Steps to reproduce:**
* WooCommerce 3.3.1
* AffiliateWP 2.1.13
* _Affiliates > Referrals > Add New_
*  Enter anything in the _Affiliates_ entry and continue filling in the referral values.  Click to **Add Referral** and nothing is saved, you are returned to the Referrals list.

pippinsplugins:
@tubiz could you look into adding improved handling here?

tubiz:
@pippinsplugins will do.

tubiz:
PR: #2577 

<img width="1271" alt="referral not created" src="https://user-images.githubusercontent.com/602184/36371644-07f49430-1563-11e8-8dbc-6208037ab771.png">

PR #2577, opened 2018-02-19
Issue/2570 - Display an error if the affiliate account is not valid when adding a new referral in the admin

tubiz:
Issue: #2570

### DIFF
--- a/includes/admin/class-notices.php
+++ b/includes/admin/class-notices.php
@@ -225,6 +225,20 @@ class Affiliate_WP_Admin_Notices {
 
 					break;
 
+				case 'referral_add_failed' :
+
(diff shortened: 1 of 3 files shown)


includes/admin/class-notices.php
```diff
@@ -225,6 +225,20 @@ public function show_notices() {
 
 					break;
 
+				case 'referral_add_failed' :
+
+					$message = __( 'Referral wasn&#8217;t created, please try again', 'affiliate-wp' );
```
DrewAPicture: We're gonna need periods at the end of these sentences :-)

tubiz: PR updated.

PR #2587, opened 2018-02-22
Added GitHub plugin URI

AiratHalitov:
Added `GitHub plugin URI` for more integration and information abilities.


AiratHalitov:
This line is needed for integration with https://github.com/afragen/github-updater plugin.

Issue #2547, opened 2018-01-31
Add affiliate referral to WooCommerce 3.3 preview order modal window


amdrew:
WooCommerce 3.3 has been released and has redesigned the orders screen. At smaller browser widths it now looks like this:

<img width="487" alt="screenshot 2018-01-31 16 41 47" src="https://user-images.githubusercontent.com/52581/35604223-21c7817c-06a6-11e8-9efe-92a4fc0e325e.png">

Note how there's no `Affiliate Referral` as outlined here: https://github.com/AffiliateWP/AffiliateWP/issues/2531#issuecomment-359965643

___

You can then click the eye icon in the screenshot above to preview the order:

<img width="452" alt="screenshot 2018-01-31 16 43 15" src="https://user-images.githubusercontent.com/52581/35604236-31f7a7e8-06a6-11e8-8e6a-c14cca5c4e8c.png">

Since you can view the affiliate referral from the main order screen in WooCommerce, and from the column shown on the orders listing screen at wider browser widths, it seems like we should add it also to this new preview order modal window.

___

You can also preview the order from the main orders screen (see icon on left). The affiliate referral column changes from https://github.com/AffiliateWP/AffiliateWP/issues/2531 on the right still look ok:

<img width="1415" alt="screenshot 2018-01-31 16 46 23" src="https://user-images.githubusercontent.com/52581/35604259-51df8c10-06a6-11e8-9d20-37a4244a63b7.png">

<img width="765" alt="screenshot 2018-01-31 16 46 36" src="https://user-images.githubusercontent.com/52581/35604270-5e332d50-06a6-11e8-9fd3-eeeb347f5949.png">

pippinsplugins:
@tubiz would you like to work on this?

tubiz:
@pippinsplugins yes.

tubiz:
PR: #2591 

<img width="513" alt="affiliate referral preview" src="https://user-images.githubusercontent.com/602184/36965194-3dcf56b6-2059-11e8-8b81-75e89ba73e49.png">

<img width="518" alt="wc order preview modal no referral" src="https://user-images.githubusercontent.com/602184/36965280-7b519a1c-2059-11e8-9244-d356f328e42f.png">

pippinsplugins:
When a referral isn't present, let's not show anything. No need for `Affiliate Referral` when there isn't one.

tubiz:
Ok. I will update the PR.

tubiz:
[...]

PR #2591, opened 2018-03-05
Issue/2547 - Add affiliate referral to WooCommerce 3.3 preview order modal window

tubiz:
Issue: #2547

### DIFF
--- a/includes/integrations/class-woocommerce.php
+++ b/includes/integrations/class-woocommerce.php
@@ -971,22 +971,16 @@ class Affiliate_WP_WooCommerce extends Affiliate_WP_Base {
 
 		$referral = affiliate_wp()->referrals->get_by( 'reference', $order_id, $this->context );
 
-		$referral_html = '<div class="wc-order-preview-affwp-referral">';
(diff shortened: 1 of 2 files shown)


includes/integrations/class-woocommerce.php
```diff
@@ -960,6 +964,41 @@ public function render_orders_referral_column( $column_name, $order_id ) {
 		}
 
 	}
+
+	public function order_preview_get_referral( $order_details, $order ) {
```
pippinsplugins: This will need PHPdoc

includes/integrations/class-woocommerce.php
```diff
@@ -960,6 +964,41 @@ public function render_orders_referral_column( $column_name, $order_id ) {
...
+
+		$referral_html .= '</div>';
+
+		$order_details['referral'] = $referral_html;
+
+		return $order_details;
+
+	}
+
+	public function render_order_preview_referral() {
```
pippinsplugins: This will need PHPdoc

tubiz: Added.

includes/integrations/class-woocommerce.php
tubiz: Added.

Issue #2374, opened 2017-09-06
Integration: OptimizeMember - referral not created on paid membership level


gspice:
Support ticket that prompted my test:
https://secure.helpscout.net/conversation/426443907/66135/?folderId=1170842

Using our sandbox site that is already configured:
http://s14890.p27.sites.pressdns.com/optimize-press

Purchased the bronze level here:
http://s14890.p27.sites.pressdns.com/optimize-press/bronze-level-stripe-form/

Using affiliate id 3:
http://s14890.p27.sites.pressdns.com/optimize-press/bronze-level-stripe-form/?ref=3

Entered Stripe test credit card, account was created and 20.00 was charged but referral was not created.   Contents of the debug log:
```
2017-9-06 18:54:20 - Visit #8 recorded for affiliate #3 in track_visit()
2017-9-06 18:55:28 - OptimizeMember payment notification.
2017-9-06 18:55:28 - OptimizeMember hash verified.
2017-9-06 18:55:28 - OptimizeMember referral validation failed. Missing affiliate ID.
2017-9-06 18:55:29 - OptimizeMember set_referral_variable() ran.
2017-9-06 18:55:29 - OptimizeMember affiliate ID was found.
2017-9-06 18:55:29 - OptimizeMember custom variable set to s14890.p27.sites.pressdns.com|3
2017-9-06 18:55:42 - OptimizeMember payment notification.
2017-9-06 18:55:42 - OptimizeMember hash verified.
2017-9-06 18:55:42 - OptimizeMember referral validation failed. Missing affiliate ID.

**Comment by gspice** [comment]
Was able to test on the user's site.  They had Fallback Tracking enable , not sure why, I didn't see any jquery errors. I tested with that on a subscription and it created a referral, actually it created two - one at a zero amount and one at the correct amount.  Here are the log entries from that testing session:
```
2017-9-06 20:33:24 - OptimizeMember set_referral_variable() ran.
2017-9-06 20:33:24 - OptimizeMember set_referral_variable() ran.
2017-9-06 20:35:11 - OptimizeMember set_referral_variable() ran.
2017-9-06 20:35:11 - OptimizeMember affiliate ID was found.
2017-9-06 20:35:11 - OptimizeMember custom variable set to www.modestycouture.com|18
2017-9-06 20:35:11 - OptimizeMember set_referral_variable() ran.
2017-9-06 20:35:11 - OptimizeMember affiliate ID was found.
[...]

PR #2611, opened 2018-03-22
Issue/2374 - Fix referral not always created in OptimizeMember integration

tubiz:
Issue: #2374


tubiz:
@pippinsplugins PR updated

PR #2622, opened 2018-04-06
issue/2623 – Allow readonly attributes in setting text fields

clickstudio:
Issue: #2623

Allow text fields to be disabled but using readonly attribute so the values can still be posted.


clickstudio:
@DrewAPicture Yes, you're right there. My bad!

See the edited file now!

DrewAPicture:
Looks good. Created #2623 for posterity.

Issue #2093, opened 2017-03-10
Admin shortcut to add an existing user as a new affiliate


DrewAPicture:
Right now there are several different pathways by which an existing user can be added as an affiliate:

1. Add a new affiliate directly via Affiliates  > Add New
2. The user can (if enabled) register themselves via the profile editor
3. Register when logged-in via the affiliate registration form
4. Via the User Migration tool (en masse)

![Add Affiliate link](https://cl.ly/1d34243A2Z1Z/Screen%20Shot%202017-03-10%20at%2010.49.37%20AM.png)

Seems to me that an obvious missing pathway is the so-called "convert this single user to an affiliate" from the admin. I propose that we add an action link to each row in the Users list table to quickly add them as a new affiliate. This would basically follow the pattern of clicking the action link, then auto-filling some values in the New Affiliate form for the given user.

Particularly interested in feedback from @sumobi and @wordmaid here.

wordmaid:
Makes sense. I like it! I'm all about the if-here-then-there. Could definitely make life a bit easier! 😄 

I think it should follow the behaviour of the other ways to add an affiliate, whereby if the **Require Approval** option is enabled in `Affiliates -> Settings -> Misc`, the affiliate goes to a _Pending_ status, and if **Require Approval** is not enabled, the affiliate goes straight to _Active_ status.

Perhaps when the Add Affiliate is successful the admin could be shown a notice notice saying `Affiliate added/created successfully. Click to Edit Affiliate.` The latter sentence could link to the Edit Affiliate screen so they can make any necessary changes (such as setting them to Active if they are set to Pending automatically)?

pippinsplugins:
I'm all for adding a link in the row actions!

sumobi:
👍

tubiz:
PR: #2652 

* Add Affiliate action link
<img width="355" alt="add affiliate action link" src="https://user-images.githubusercontent.com/602184/39518035-8fb733ae-4df9-11e8-8bc8-1be6ed976311.png">

* Add existing user as a new affiliate
![add user as affiliate](https://user-images.githubusercontent.com/602184/39517928-4a1730ce-4df9-11e8-9464-4612adfbf956.png)

pippinsplugins:
PR looks good, we can merge after 2.2 is released.

pippinsplugins:
@michaelbeil could you please test this when back from WCEU?

michaelbeil:
[...]

PR #2652, opened 2018-05-02
Issue/2093 - Admin shortcut to add an existing user as a new affiliate

tubiz:
Issue: #2093

### DIFF
--- a/includes/admin/affiliates/new.php
+++ b/includes/admin/affiliates/new.php
@@ -1,6 +1,8 @@
 <?php
 $default_rate = affiliate_wp()->settings->get( 'referral_rate', 20 );
 $default_rate = affwp_abs_number_round( $default_rate );
+$user_id      = isset( $_GET['user_id'] ) ? absint( $_GET['user_id'] ) : '';
+$user         = ! empty( $user_id ) && ! affwp_is_affiliate( $user_id ) ? get_userdata( $user_id ) : '';
 ?>
 <div class="wrap">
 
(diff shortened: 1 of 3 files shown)


includes/admin/affiliates/new.php
```diff
@@ -65,14 +77,25 @@
...
 
-				<td>
-					<select name="status" id="status" disabled="disabled">
-						<option value="active"><?php _e( 'Active', 'affiliate-wp' ); ?></option>
-						<option value="inactive"><?php _e( 'Inactive', 'affiliate-wp' ); ?></option>
-						<option value="pending"><?php _e( 'Pending', 'affiliate-wp' ); ?></option>
-					</select>
-					<p class="description"><?php _e( 'The status assigned to the affiliate&#8217;s account.', 'affiliate-wp' ); ?></p>
-				</td>
+				<?php if ( $user ): ?>
```
pippinsplugins: I don't understand the purpose of these `if( $user )` checks. Could you explain @tubiz ?

tubiz: @pippinsplugins when adding a non-registered user the Affiliate Status select field has a disabled attribute until a user is selected. 
Thie check is to remove the disable attribute to allow the Affiliate Status to be selected when adding an already registered user.

pippinsplugins: Ah, perfect.

Could we reduce this, however, so that the HTML is not duplicated? Really only the disabled attribute should be inside of the conditional statement.

tubiz: PR updated. I noticed some additional duplicated HTML which has also been fixed.

Issue #2655, opened 2018-05-03
Formidable Pro referrals are generated with a status of pending


michaelbeil:
When you activate the Formidable Pro pro integration, referrals are generated with a status of pending. These should be accepted and set to a status of unpaid.

![pnnmd6](https://user-images.githubusercontent.com/4752949/39577649-33777858-4ee2-11e8-95bb-10ef23f54cdc.png)

Ticket: https://secure.helpscout.net/conversation/573085450/81503?folderId=205524.

tubiz:
@pippinsplugins I will need a copy of the Formidable Pro plugin.

pippinsplugins:
@tubiz Formidable is free on WordPress.org. You should not need to Pro version to replicate.

tubiz:
@pippinsplugins I was able to replicate the issue using the Formidable plugin on WordPress.org

A referral is mark as complete using the `frm_after_payment_completed` hook which isn't present in the Formidable plugin but in any the Formidable payment add-ons.

pippinsplugins:
Maybe we should go ahead and introduce an option for the default referral status on a form then? That could work well after 2.2 is released when we have referral types for `lead` and `opt-in` added.

wilkdavis:
(not really a github user so please forgive me if jumping into your thread here is impolite).  
It looks like that frm_after_payment_completed hook is no longer used.  I sent the following over to Formidable and got the response below...
----me, to Formidable----
Github issue comment: 
"A referral is mark as complete using the frm_after_payment_completed"

(Mostly just curious and trying to follow along and learn something), I search for that and only find it in 2 places - a 2yr old ticket on your site, and here:
https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/integrations/class-formidablepro.php

Which makes me wonder... should that be using something like these instead? 

frm_payment_status_complete
frm_payment_status_failed
frm_payment_status_refunded

(and if so, do these work with any of the payment gateways, or does each payment addon use different hooks (like if using paypal, would one have to use frm_payments_paypal_ipn?)

---response from Formidable----
"I've been researching to find where that hook originated. It looks like the frm_after_payment_completed is used in a 3rd-party Stripe addon that is no longer supported. You're exactly right that our Stripe and Authorize.net addons use those other three hooks you listed above.

We have plans of rewriting the PayPal addon to use the same base as Stripe and Authorize.net. It would definitely make sense to add those three hooks into PayPal as well. At the moment, the
[...]

PR #2676, opened 2018-05-21
Issue/2655 - Fix issue with completing and revoking referrals in Formidable Pro integration

tubiz:
Issue: #2655


tubiz:
PR updated with the fix for marking referral as complete via the PayPal payment add-on

Issue #2681, opened 2018-05-23
Database error: Column 'email' cannot be null for query INSERT INTO wp_affiliate_wp_customers


tubiz:
When a referral is created via the Contact Form 7 and Formidable Pro integrations the database error below is shown, due to the customer email not present.

```
[23-May-2018 09:19:42 UTC] WordPress database error Column 'email' cannot be null for query INSERT INTO `wp_affiliate_wp_customers` (`date_created`, `first_name`, `last_name`, `email`, `user_id`, `ip`) VALUES ('2018-05-23 09:19:42', '', '', NULL, 0, '192.168.95.1') made by require('wp-blog-header.php'), wp, WP->main, WP->parse_request, do_action_ref_array, WP_Hook->do_action, WP_Hook->apply_filters, call_user_func_array, wpcf7_control_init, WPCF7_ContactForm->submit, do_action('wpcf7_submit'), WP_Hook->do_action, WP_Hook->apply_filters, call_user_func_array, Affiliate_WP_Contact_Form_7->add_pending_referral, Affiliate_WP_Base->insert_pending_referral, Affiliate_WP_Referrals_DB->add, Affiliate_WP_Referrals_DB->setup_customer, Affiliate_WP_Customers_DB->add, Affiliate_WP_DB->insert
```

tubiz:
PR: #2682

PR #2682, opened 2018-05-23
Issue/2681  - Only add an affiliate customer record if the customer email is set

tubiz:
Issue: #2681


includes/class-customers-db.php
```diff
@@ -343,6 +343,10 @@ public function add( $args = array() ) {
 
 		$args = wp_parse_args( $args, $defaults );
 
+		if( empty( $args['email'] ) ) {
+			return false;
```
DrewAPicture: We should log an error to the debug log if there's no email instead of just returning false. The debug log can help users figure out why something isn't working.

DrewAPicture: Also, if we're adding an early return, we'll need a test to cover it :-)

pippinsplugins: I agree with @DrewAPicture

tubiz: I will update the PR with the requested changes.

tubiz: @DrewAPicture PR updated.

Issue #2684, opened 2018-05-24
Disable IP address logging


michaelbeil:
We should add an option to disable IP address logging.  

Ticket: https://secure.helpscout.net/conversation/586272751/82700/.

pippinsplugins:
@tubiz could you put this at the top of your list to finish?

tubiz:
PR: #2690

michaelbeil:
Related #2680.

michaelbeil:
An IP address is not logged with the setting enabled to disable IP address logging in the attached PR.

pippinsplugins:
@tubiz please see the one change request on the PR.

PR #2690, opened 2018-05-25
Issue/2684 - Disable IP address logging

tubiz:
Issue: #2684

### DIFF
--- a/includes/class-tracking.php
+++ b/includes/class-tracking.php
@@ -860,15 +860,23 @@ class Affiliate_WP_Tracking {
 	 * @since 1.0
 	 */
 	public function get_ip() {
-		if ( ! empty( $_SERVER['HTTP_CLIENT_IP'] ) ) {
-			//check ip from share internet
-			$ip = $_SERVER['HTTP_CLIENT_IP'];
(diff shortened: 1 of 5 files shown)


includes/class-tracking.php
```diff
@@ -860,15 +860,23 @@ public function is_valid_affiliate( $affiliate_id = 0 ) {
...
-			$ip = $_SERVER['HTTP_CLIENT_IP'];
-		} elseif ( ! empty( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) {
-			//to check ip is pass from proxy
-			$ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
-		} else {
-			$ip = $_SERVER['REMOTE_ADDR'];
+
+		$ip = '';
+
+		if ( ! affiliate_wp()->settings->get( 'disable_ip_logging' ) ) {
```
pippinsplugins: I don't think this should be included. The function to retrieve an IP should still work as there may be cases where the IP needs to be detected but simply not stored. Let's leave this function as is.

tubiz:
@pippinsplugins PR updated.

Issue #2726, opened 2018-06-17
Clearing out terms label should show default text


amdrew:
As discussed in https://github.com/AffiliateWP/AffiliateWP/issues/2713#issuecomment-397908622, clearing out the terms label should not show a blank field. This is confusing, since the text is blank in the admin, yet the front-end shows text:

<img width="1651" alt="screen shot 2018-06-18 at 9 40 51 am" src="https://user-images.githubusercontent.com/52581/41512490-195e7be6-72dd-11e8-8c08-94efd3f2d94d.png">

Instead, when the field is cleared out, it should show the default text specified in the codebase: https://github.com/AffiliateWP/AffiliateWP/blob/master/includes/admin/settings/class-settings.php#L487

This has the added bonus of showing the translated text in another language, if it's available.

<img width="1671" alt="screen shot 2018-06-18 at 9 41 15 am" src="https://user-images.githubusercontent.com/52581/41512493-23bb04d8-72dd-11e8-8d6b-edc530f08a53.png">

tubiz:
PR: #2728

gspice:
Thank you for this change @amdrew and @tubiz .  This makes it a lot easier to understand working with, and creating translation changes.

PR #2728, opened 2018-06-18
Issue/2726 - Always use the default text in the text_callback() method in the Affiliate_WP_Settings class if the text field is empty.

tubiz:
Issue: #2726


amdrew:
@tubiz @pippinsplugins 

Looks good to me 👍I tested the other text settings fields and they all work great too.

Issue #2772, opened 2018-07-23
Customer record not created in RCP integration for customers that are not logged in 


tubiz:
When a referral is being created via the RCP integration for a customer that is not logged in, the customer record is not created.

```
2018-7-23 10:01:50 - Arguments being sent to DB: array (
  'amount' => '5',
  'reference' => 'fc759c64e476d97ab2891fbd6a17fc6a',
  'description' => 'RCP Level A',
  'campaign' => '',
  'affiliate_id' => 2,
  'visit_id' => 51,
  'products' => '',
  'custom' => '',
  'type' => 'sale',
  'context' => 'rcp',
  'customer' => 
  array (
    'first_name' => '',
    'last_name' => '',
    'email' => NULL,
    'user_id' => 0,
    'ip' => '127.0.0.1',
    'affiliate_id' => 2,
  ),
)
2018-7-23 10:01:50 - Customer not created due to empty email
2018-7-23 10:01:50 - Pending Referral #139 created successfully.
2018-7-23 10:01:59 - Referral retrieved successfully during complete_referral()
2018-7-23 10:01:59 - Referral #139 set to Unpaid successfully
```

Related: https://github.com/AffiliateWP/affiliate-wp-lifetime-commissions/issues/70

tubiz:
PR: #2773 

The `rcp_subscription_key` usermeta is not yet saved as at the time it is retrieved so the `user_id` is always empty. The `user_id` is now retrieved from the `rcp_payments` table.

michaelbeil:
Is it best to use an email for a customer (already referred by an affiliate) that is logged out for a purchase made with RCP to test this?

tubiz:
@michaelbeil to test, use a non registered user. You will have to enter the new user details on the RCP registration page.

michaelbeil:
Okay, perfect. I just tested with a non-registered user, and a referral was successfully generated for both the initial referral and then the second referral when logged in with the customer account:
 
![screen shot on 2018-07-23 at 18_43_07](https://user-images.githubusercontent.com/4752949/43090673-588225de-8ea8-11e8-9432-eb837f2a15e4.png)

tubiz:
@michaelbeil you are to test if the customer record was created. From the debug log the customer is not created and linked to the affiliate due to this error

`2018-7-23 10:01:50 - Customer not created due to empty email`

PR #2773, opened 2018-07-23
Issue/2772 - Fix customer record not created in RCP integration for customers that are not logged in

tubiz:
Issue: #2772

### DIFF
--- a/includes/integrations/class-rcp.php
+++ b/includes/integrations/class-rcp.php
@@ -186,8 +186,8 @@ class Affiliate_WP_RCP extends Affiliate_WP_Base {
 				$user = get_userdata( $user_id );
 
 				$customer = array(
-					'first_name'   => $user ? $user->last_name : '',
-					'last_name'    => $user ? $user->first_name : '',
+					'first_name'   => $user ? $user->first_name : '',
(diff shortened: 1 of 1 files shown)


includes/integrations/class-rcp.php
```diff
@@ -200,7 +202,7 @@ public function get_customer( $subscription_key = 0 ) {
 			$customer = array(
 				'first_name'   => is_user_logged_in() ? wp_get_current_user()->last_name : '',
 				'last_name'    => is_user_logged_in() ? wp_get_current_user()->first_name : '',
-				'email'        => is_user_logged_in() ? wp_get_current_user()->user_email : $this->email,
+				'email'        => is_user_logged_in() ? wp_get_current_user()->user_email : '',
```
pippinsplugins: Could this look to see if the post data contains the email and fall back to that?

tubiz: Yes. I will update the PR

tubiz: The `post_data` contains the `email` when the user is not logged in, but the `email` property was deprecated in `2.2`

